PR #22, opened 2015-01-06
Generate request id for secondary watcher log.

chitang:


coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -37,6 +38,7 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
 
     val foundWorkToDo = for (job <- secondaryManifest.claimDatasetNeedingReplication(
                                       secondary.storeId, claimantId, claimTimeout)) yield {
+      MDC.put(RequestId.ReqIdHeader, RequestId.generate())
```
znep: Don't forget to clear it out as well just to be hygienic.

PR #23, opened 2015-01-06
Make secondary watcher log easier to read by attaching a unique id and d...

chitang:
...ataset id for each job.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
@@ -6,8 +6,8 @@ import com.socrata.datacoordinator.common.DataSourceFromConfig.DSInfo
 import com.socrata.datacoordinator.common.{SoQLCommon, DataSourceFromConfig}
 import com.socrata.datacoordinator.secondary.sql.SqlSecondaryConfig
 import com.socrata.datacoordinator.truth.universe._
-import com.socrata.datacoordinator.util.{TimingReport, NullCache, StackedTimingReport, LoggedTimingReport,
-                                         MetricsTimingReport}
+import com.socrata.datacoordinator.util._
+import com.socrata.http.server.util.RequestId
 import com.socrata.soql.types.{SoQLValue, SoQLType}
 import com.socrata.thirdparty.metrics.{MetricsOptions, MetricsReporter}
 import com.socrata.thirdparty.typesafeconfig.Propertizer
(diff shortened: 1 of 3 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -26,7 +26,7 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
                                claimantId: UUID,
                                claimTimeout: FiniteDuration,
                                timingReport: TimingReport) {
-  import SecondaryWatcher.log
+  import SecondaryWatcher._
```
kaidamasaki: Switching `SecndaryWatcher.log` to `SecondaryWatcher._` seems weird here. Are you pulling in a lot of things from there?

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -37,26 +37,31 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
         try {
-          secondaryManifest.releaseClaimedDataset(job)
+          playbackToSecondary(secondary, job)
+          log.info("<< Sync done for {} into {}", job.datasetId, secondary.storeId)
         } catch {
           case e: Exception =>
-            log.error("Unexpected exception while releasing claim on dataset {} in secondary {}",
-                      job.datasetId.asInstanceOf[AnyRef], secondary.storeId, e)
+            log.error("Unexpected exception while updating dataset {} in secondary {}; marking it as broken",
+              job.datasetId.asInstanceOf[AnyRef], secondary.storeId, e)
```
kaidamasaki: Maybe match old indentation?  (Indent to beginning of paren.)

coordinatorlib/src/main/scala/com/socrata/datacoordinator/util/TimingReport.scala
```diff
@@ -69,6 +69,36 @@ class LoggedTimingReport(log: Logger) extends TimingReport {
...
+ */
+trait TaggableTimingReport extends TimingReport with Metrics {
+
+  private val Tag = "^tag:(.+)$".r
+
+  abstract override def apply[T](name: String, kv: (String, Any)*)(f: => T): T = {
+
+    val (regKv, tagKv) = kv.partition { case (tag, value) => Tag.findFirstIn(tag).isEmpty }
+
+    val tagK = tagKv.collect { case (tag, value) =>
```
kaidamasaki: Could this be folded to `case (Tag(tag), value)`?  Or does the compiler not like that?  (I guess or do you want an error to be thrown if `tag` isn't a `Tag`.)

PR #24, opened 2015-01-15
Reduce metric clutter.  There is also a little bit of formatting.

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryConfig.scala
```diff
@@ -10,22 +10,26 @@ import com.socrata.datacoordinator.util.TimingReport
...
+    val sql = """
+      SELECT store_id, next_run_time, interval_in_seconds
+        FROM secondary_stores_config
+       WHERE store_id = ?""".stripMargin
+
+    for {
+      stmt <- managed(conn.prepareStatement(sql))
+      _ <- unmanaged(stmt.setString(1, storeId))
+      rs <- managed(stmt.executeQuery())
+    } yield {
```
sokrahta: :+1: nice use of for-comprehension

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryConfig.scala
```diff
@@ -10,22 +10,26 @@ import com.socrata.datacoordinator.util.TimingReport
...
+      if (rs.next())
+        Some(SecondaryConfigInfo(
+          rs.getString("store_id"),
+          new DateTime(rs.getTimestamp("next_run_time").getTime),
+          rs.getInt("interval_in_seconds")))
+      else
+        None
```
sokrahta: style: use braces around if/else blocks

sokrahta:
there's no test coverage on this method, just saying.
        <method name="com.socrata.datacoordinator.secondary.sql/SqlSecondaryConfig/lookup" signature="()V" line-rate="0.00" branch-rate="0.00">

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryConfig.scala
kaidamasaki: One could argue not using braces is correct here.  
Since this is an `if` expression, not an `if` statement.

Though braces could potentially make it clearer.

velvia: I think it's still better to use braces.  Consistency, and no matter if it's an expression it's multi line and adding an extra line to the else would easily turn into a silent error.

velvia:
Those two metrics seem harmless to remove.

znep:
I agree that these particular ones are fine to remove, but it does bring up
a larger goal... it is really nice have a system where you can instrument
calls by default and not be worried about them being "clutter".  Timing
information you can actually query on allows you to trivially filter out
fast thing, aggregate fast things by count which may make them more
significant, etc.

sokrahta:
retest this please

PR #28, opened 2015-03-11
Clean up a few things that relate only to code that has been previously removed.

znep:


velvia:
Why is none of this stuff needed anymore?

znep:
Because you removed the code that used it in 3d0150f7c556ea5d2905605dbbb29dc39d2357dc

velvia:
Oh right.

PR #30, opened 2015-03-23
Add docker support.

znep:


urmilan:
There should be a run script for secondary watcher as well, no? Apart from that looks good.

znep:
Yup, leaving that for a separate step since it requires "more stuff".  It
needs its whole own separate setup, secondary jars, secondary database info.

urmilan:
Okie dokie.

PR #31, opened 2015-03-26
Fix secondary groups so we actually ensure we have the desired number of copies in each group.

znep:
We were only checking to make sure we have the right number of copies total, making secondary groups not
work at all.

I wrote this over a year ago with the exact use case I have now in mind, only we haven't used it yet
and I failed on writing tests because I could barely write Scala so it didn't work.  The test now is
weak, but I didn't want to refactor the whole class right now.

The only intended actual code change is adding "intersect(secondaryGroup.instances)", everything else is
just to get the logic in a place it can be easily tested.


velvia:
LGTM overall (besides nice to add a comment).  

Slightly related note, how do we detect that data in different replicas are out of sync?

I wonder if long term we should consider something like pg_shard.  I believe it takes care of replication at the very least, and I believe data synchronicity as well.

velvia:
LGTM, merging.

PR #33, opened 2015-03-29
Grr, typesafe config won't use relative paths apparently.

znep:


velvia:
I think absolute paths are safer anyways.

PR #39, opened 2015-04-12
effectively unlimited query pool for secondary watcher

znep:
Running out of pooled connections can result in strange and difficult to detect failure conditions. For example, the claim manager can get blocked resulting in claims not being updated and thus stolen. We control concurrency by the number of workers combined with the number of secondary instances configured, so we set it to a "essentially unlimited" value for the secondary watcher.


chitang:
Can we put make a TODO note to fix this too?

znep:
Sure, done.

PR #40, opened 2015-04-13
Cap DC memory usage - limit # mutation threads

velvia:
Each mutation thread can use a significant amount of memory due to batching up mutations before writing to truth.  Cap the number of mutation threads allowed using an AtomicInteger.  Note this number should be smaller than the total number of threads serving requests to allow for exports and other requests.  

Also, reduce the default batch size.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
@@ -233,22 +233,21 @@ class Service(serviceConfig: ServiceConfig,
       false
     } else {
       val currThreads = numThreads.get()
-      if (currThreads >= serviceConfig.maxMutationThreads) {
-        log.info(s"tryGetMutationThread: too many threads ($currThreads), waiting for some to finish")
+      if (currThreads >= serviceConfig.maxMutationThreads ||
+          !numThreads.compareAndSet(currThreads, currThreads + 1)) {
+        if (currThreads >= serviceConfig.maxMutationThreads)
+          log.info(s"tryGetMutationThread: too many threads ($currThreads), waiting for some to finish")
         Thread sleep 100
         tryGetMutationThread(numTries - 1)
-      } else if (numThreads.compareAndSet(currThreads, currThreads + 1)) {
-        true
       } else {
-        Thread sleep 100
-        tryGetMutationThread(numTries - 1)
+        true
       }
     }
   }
 
   def withMutationScriptResults[T](f: => HttpResponse): HttpResponse = {
     if (!tryGetMutationThread()) {
-      return err(ServiceUnavailable, "mutation.threads.maxed.out")
+      return err(ServiceUnavailable, "mutation.threads.maxed-out")
     }
     try {
       f
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
@@ -227,9 +227,9 @@ class Service(serviceConfig: ServiceConfig,
   // Repeatedly tries to get a new thread by bumping numThreads, using
   // compareAndSet to make sure our value is valid
   @annotation.tailrec
-  private def tryGetMutationThread(numTries: Int = 100): Boolean = {
-    if (numTries <= 0) {
-      log.info("tryGetMutationThread: out of tries")
+  private def tryGetMutationThread(timeoutMs: Long, waitInterval: Int = 100): Boolean = {
+    if (timeoutMs <= 0) {
+      log.info("tryGetMutationThread: timed out")
       false
     } else {
       val currThreads = numThreads.get()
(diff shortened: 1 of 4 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -221,7 +224,32 @@ class Service(serviceConfig: ServiceConfig,
...
+      } else {
+        Thread sleep 100
+        tryGetMutationThread(numTries - 1)
+      }
+    }
+  }
+
   def withMutationScriptResults[T](f: => HttpResponse): HttpResponse = {
+    if (!tryGetMutationThread()) {
+      return err(ServiceUnavailable, "mutation.threads.maxed.out")
```
znep: Nit: should be "maxed-out", since I don't think we would have another value like "maxed.in".  Periods are hierarchy.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -221,7 +224,32 @@ class Service(serviceConfig: ServiceConfig,
       "dataset" -> JString(formatDatasetId(name)),
       "schema" -> jsonifySchema(schema))
 
+  // Repeatedly tries to get a new thread by bumping numThreads, using
+  // compareAndSet to make sure our value is valid
+  @annotation.tailrec
+  private def tryGetMutationThread(numTries: Int = 100): Boolean = {
```
znep: So if I'm reading this right we end up sleeping a maximum of 10 seconds.  That seems like a pretty small amount of time given how relatively heavyweight mutations are by nature and how failing too quickly can cause unnecessary pain on retries.  

It seems like something on the order of a few minutes might be more reasonable?  Especially if combined with setting a max jetty thread pool as a failsafe, it doesn't look like we have done that yet for dc unless I missed it.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -221,7 +224,32 @@ class Service(serviceConfig: ServiceConfig,
...
+    if (numTries <= 0) {
+      log.info("tryGetMutationThread: out of tries")
+      false
+    } else {
+      val currThreads = numThreads.get()
+      if (currThreads >= serviceConfig.maxMutationThreads) {
+        log.info(s"tryGetMutationThread: too many threads ($currThreads), waiting for some to finish")
+        Thread sleep 100
+        tryGetMutationThread(numTries - 1)
+      } else if (numThreads.compareAndSet(currThreads, currThreads + 1)) {
```
znep: Can we combine these two conditions with sleeps/recurstion into a single (currThreads >= serviceConfig.maxMutationThreads || !numThreads.compareAndSet(currThreads, currThreads + 1)?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractRepBasedDataSqlizer.scala
```diff
@@ -11,8 +11,17 @@ import com.socrata.datacoordinator.id.{ColumnId, RowId}
 import com.socrata.datacoordinator.util.collection.{ColumnIdSet, RowIdSet, MutableRowIdSet}
 import java.io.Closeable
 
+/**
+ * Abstract SQL adapter for data access
+ *
+ * @param softMaxBatchSize in bytes, the max size of each batch to insert.
+ * This is actually very tricky because the size is estimated from the SqlRep's estimatedSize
+ * method for each column value, but doesn't include the intermediate objects produced, which
+ * are actually the most expensive and may take 100x as much memory as the estimated size.
```
znep: I haven't looked at this at all so I have no idea what the answer is, but I assume there is no easy way to refactor to better take this into account?  Not saying we need to or should now.

znep:
Reviewed up to 6f05cc.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractRepBasedDataSqlizer.scala
rjmac: It's only vaguely "in bytes"; the intent isn't to give an exact (or even really very rough) value so much as a vague "this is big, this is small, we probably have a lot of data and should flush it to the DB now" estimate.

Maybe it should be `@param softMaxBatchSize in story points`

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
velvia: True.   I'll change this to accept a duration, and to make it configurable.   Maybe 5 minutes default?

As for the max Jetty thread pool, we should do that too, but possibly another PR.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractRepBasedDataSqlizer.scala
velvia: hahaha.

velvia:
Addressed review comments.

PR #43, opened 2015-04-24
Support for getting info on existing rollups.

znep:


coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -238,7 +238,17 @@ object Main {
           }
         }
 
-        val serv = new Service(serviceConfig, operations.processMutation, operations.processCreation, getSchema _,
+        def getRollups(datasetId: DatasetId) = {
```
velvia: Shouldn't this be collocated in Service.scala?

velvia:
Looks fine for the most part.  Reviewed up to 56d798c

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
znep: Yeah... that is a hard question to answer.  For now I'm just trying to follow the same patterns as exist for, say, getSchema.  For better or worse.

PR #48, opened 2015-05-08
Cannot run migration in docker process because the docker config is m…

chitang:
…is-aligned.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/primary/MigrateSchema.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/primary/MigrateSchema.scala
@@ -37,5 +37,6 @@ object MigrateSchema extends App {
 
     SchemaMigrator(databaseTree, operation, numChanges)
   }
-  private lazy val databaseTree = "com.socrata.coordinator.common.database"
+
+  private lazy val databaseTree = s"${com.socrata.datacoordinator.service.Main.configRoot}.database"
 }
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/primary/MigrateSchema.scala
```diff
@@ -37,5 +37,6 @@ object MigrateSchema extends App {
 
     SchemaMigrator(databaseTree, operation, numChanges)
   }
-  private lazy val databaseTree = "com.socrata.coordinator.common.database"
+
+  private lazy val databaseTree = s"${com.socrata.datacoordinator.service.Main.configRoot}.database"
```
velvia: Um, you are saying there is a variable named `com.socrata.datacoordinator.service.Main.configRoot` ?  Are you sure?

chitang: It is more like a constant.  Is it the name, being a val or relating the two classes that concerns you?

There are alternatives like
1. repeat the same string literal.
2. change the config file.
3. make Make.make run the migration class by detecting some argument.
and more...

velvia: Oh. I see.  It is the fully qualified variable name including package.  That's fine, but slightly confusing.  I think it would be better if you imported Main and just did Main.configRoot, or something similar.

sokrahta:
lgtm :shipit:
(except :person_frowning: no tests covering the changes)

PR #50, opened 2015-05-15
Refs CORE-5412

rozap:
* added new types to SoqlRep and RowLogCodecs
* change migrate loglevel to supress some spam
* todo: update dependencies file and remove snapshot ref

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/migration/Migration.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/migration/Migration.scala
@@ -2,6 +2,7 @@ package com.socrata.datacoordinator.truth.migration
 
 import liquibase.Liquibase
 import liquibase.database.jvm.JdbcConnection
+import liquibase.logging.LogFactory
 import liquibase.resource.ClassLoaderResourceAccessor
 
 import java.sql.Connection
@@ -26,7 +27,7 @@ object Migration {
                 operation: MigrationOperation = MigrationOperation.Migrate,
                 numChanges: Int = 1,
                 changeLogPath: String = MigrationScriptPath) {
-
+    LogFactory.getLogger().setLogLevel("warning")
     val liquibase = new Liquibase(changeLogPath, new ClassLoaderResourceAccessor, new JdbcConnection(conn))
     val database = conn.getCatalog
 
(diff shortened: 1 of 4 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/migration/Migration.scala
```diff
@@ -26,7 +27,7 @@ object Migration {
                 operation: MigrationOperation = MigrationOperation.Migrate,
                 numChanges: Int = 1,
                 changeLogPath: String = MigrationScriptPath) {
-
+    LogFactory.getLogger().setLogLevel("warning")
```
urmilan: What's this for?

urmilan: Nvm, just saw the explanation in the PR comment.

chitang: I like to cut down logging.  But I find it helpful to have INFO in migration.  Feel better w/o going to check the database if and which migration is done.

rozap: sounds good, i'll remove those

rozap:
google isn't helping much. is there a way to specify the loglevel based on environment? for example, running `sbt test` is a test environment, so it would be helpful to bump up the log level so test output is shown, whereas in a regular dev environment the loglevel would be lower.

velvia:
Specifying the log level based on env is typically done by changing the log
configuration.  For a lot of our services the log config comes from the
Typesafe config file.  That way, you can use deploy tools to change what
logging config is used.

For other services that rely on log4j, you'd just modify or deploy a
different log4j.properties.

On Fri, May 15, 2015 at 3:48 PM, Chris Duranti <notifications@github.com>
wrote:

> google isn't helping much. is there a way to specify the loglevel based on
> environment? for example, running sbt test is a test environment, so it
> would be helpful to bump up the log level so test output is shown, whereas
> in a regular dev environment the loglevel would be lower.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/socrata-platform/data-coordinator/pull/50#issuecomment-102542356>
> .
>



-- 
Evan Chan   |   Principal Engineer, Socrata     |    @evanfchan
*Socrata <http://www.socrata.com/> *| The Data Experience Company™
*Socrata isn't about worldly success, but about changing the world
successfully. This is why I'm excited to be here.*
*What does Socrata do? <https://www.youtube.com/watch?v=yyz86MlkriQ> *(watch
our 2 min video)

urmilan:
If you've resolved @velvia and @chitang's concerns, LGTM. Remember to replace snapshot dependencies before merging

PR #52, opened 2015-06-02
[CORE-5319] Auto retry on replication failure

velvia:
This PR adds functionality to automatically retry on failure, with exponential back off.  More precisely, on failure, the dataset is marked broken only if the # of retries has not been exceeded; otherwise the retry information is updated in the secondary_manifest table, and the watcher can pick up the job again.

The back off interval and max number of retries can be configured in the secondary watcher .conf file.

Also, a test for the secondary watcher has been added.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
@@ -26,6 +26,8 @@ import sun.misc.{Signal, SignalHandler}
 class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseType[CT, CV]],
                                claimantId: UUID,
                                claimTimeout: FiniteDuration,
+                               backoffInterval: FiniteDuration,
+                               maxRetries: Int,
                                timingReport: TimingReport) {
   val log = LoggerFactory.getLogger(classOf[SecondaryWatcher[_,_]])
   private val rand = new Random()
@@ -51,14 +53,23 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
         "endingDataVersion" -> job.endingDataVersion
       ) {
         log.info(">> Syncing {} into {}", job.datasetId, secondary.storeId)
+        if (job.retryNum > 0) log.info("Retry #{} of {}", job.retryNum, maxRetries)
         try {
           playbackToSecondary(secondary, job)
           log.info("<< Sync done for {} into {}", job.datasetId, secondary.storeId)
         } catch {
           case e: Exception =>
-            log.error("Unexpected exception while updating dataset {} in secondary {}; marking it as broken",
-                      job.datasetId.asInstanceOf[AnyRef], secondary.storeId, e)
-            getManifest(u).markSecondaryDatasetBroken(job)
+            if (job.retryNum < maxRetries) {
+              val retryBackoff = backoffInterval.toSeconds * Math.pow(2, job.retryNum)
+              log.warn("Unexpected exception while updating dataset {} in secondary {}, retrying in {}...",
+                       job.datasetId.asInstanceOf[AnyRef], secondary.storeId, retryBackoff.toString, e)
+              getManifest(u).updateRetryInfo(secondary.storeId, job.datasetId, job.retryNum + 1,
+                                             retryBackoff.toInt)
+            } else {
+              log.error("Unexpected exception while updating dataset {} in secondary {}; marking it as broken",
+                        job.datasetId.asInstanceOf[AnyRef], secondary.storeId, e)
+              getManifest(u).markSecondaryDatasetBroken(job)
+            }
         } finally {
           try {
             getManifest(u).releaseClaimedDataset(job)
@@ -227,7 +238,8 @@ object SecondaryWatcher extends App { self =>
       NullCache
     )
 
-    val w = new SecondaryWatcher(common.universe, config.watcherId, config.claimTimeout, common.timingReport)
+    val w = new SecondaryWatcher(common.universe, config.watcherId, config.claimTimeout,
+                                 config.backoffInterval, config.maxRetries, common.timingReport)
     val cm = new SecondaryWatcherClaimManager(dsInfo, config.watcherId, config.claimTimeout)
 
     val SIGTERM = new Signal("TERM")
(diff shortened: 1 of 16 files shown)


velvia:
@znep done.

znep:
Oh, sorry I should have been more specific... we have this index already:

    "secondary_manifest_order" btree (store_id, (broken_at IS NULL),
(latest_data_version > latest_secondary_data_version), went_out_of_sync_at)

That is used to find datasets needing replication.  We should add the
next_retry condition to that one, and verify with explain that the new
query can be satisfied from the index.

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20150601-add-column-next-retry.xml
```diff
@@ -0,0 +1,24 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="evan" id="20150601-add-column-next-retry">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="secondary_manifest" columnName="next_retry"/>
+            </not>
+        </preConditions>
+        <sql>
+            ALTER TABLE secondary_manifest ADD next_retry TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
znep: Should this be `timestamp with time zone` to match the other columns?

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -26,17 +26,23 @@ import sun.misc.{Signal, SignalHandler}
...
+                               maxRetries: Int,
                                timingReport: TimingReport) {
-  import com.socrata.datacoordinator.secondary.SecondaryWatcher.log
+  val log = LoggerFactory.getLogger(classOf[SecondaryWatcher[_,_]])
   private val rand = new Random()
   // splay the sleep time +/- 5s to prevent watchers from getting in lock step
   private val nextRuntimeSplay = (rand.nextInt(10000) - 5000).toLong
 
+  // allow for overriding for easy testing
+  protected def getManifest(u: Universe[CT, CV] with SecondaryManifestProvider with PlaybackToSecondaryProvider):
```
sokrahta: Scala getters conventionally don't start with the word `get`

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -47,17 +53,26 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
+        if (job.retryNum > 0) log.info("Retry #{} of {}", job.retryNum, maxRetries)
         try {
           playbackToSecondary(secondary, job)
           log.info("<< Sync done for {} into {}", job.datasetId, secondary.storeId)
         } catch {
           case e: Exception =>
-            log.error("Unexpected exception while updating dataset {} in secondary {}; marking it as broken",
-                      job.datasetId.asInstanceOf[AnyRef], secondary.storeId, e)
-            secondaryManifest.markSecondaryDatasetBroken(job)
+            if (job.retryNum < maxRetries) {
```
sokrahta: maxRetries can be omitted considering that exponential backoff will quickly put the next retry time far enough in the future to be irrelevant.

coordinator/src/test/scala/com/socrata/datacoordinator/secondary/SecondaryWatcherTest.scala
```diff
@@ -0,0 +1,105 @@
...
+      // Mock a secondary, set up expectations
+      val testSecondary = mock[Secondary[common.CT, common.CV]]
+
+      (testManifest.updateRetryInfo _).expects(testStoreId, datasetId, 1, 10)
+
+      // Run the watcher run() method
+      w.run(u, new NamedSecondary(testStoreId, testSecondary))
+    }
+  }
+}
```
sokrahta: :+1: Hurrah for tests!

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20150601-add-column-next-retry.xml
velvia: Yep.  Good catch.

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
velvia: I would disagree.  This would result in datasets not being marked broken, ever...  I'd rather have a measure of finality.   We can have a max on the total time for retrying.   Also, with the default back off interval, it would take 10 retries for the backup to be a day.

sokrahta: This pr sets broken after a number of retries, but it seems reasonable to set broken after the first failure.

And with a retry backoff of 1 minute and exponential growth of 2^n sure it will take that long, but how often do we want to retry and how quickly to give up? With the proposed values it will be done retrying after only 14 minutes.

sokrahta:
build issue: 1 test failed
> 
[info] TestSqlLoader:
[info] - building a report before finished fails (simple report writer) (53 milliseconds)
[info] - adding a new row with system PK succeeds *** FAILED *** (167 milliseconds)
[info]   java.sql.SQLException: No suitable driver found for jdbc:h2:mem:

velvia:
@znep @sokrahta @chitang Most of your feedback is incorporated.

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
velvia: Well keep in mind most replications take many many minutes, often hours, sometimes more than a day.   The way the timing works is that after the first failure, next_retry is set to some time in the future.  The watcher loops back, it might not even take that same item up, some other item might have moved up the stack.  Thus the back off actually won't do that much.....   unless the failures occur right away.

We could certainly increase the defaults ... perhaps to 5 minutes * 5 retries = max back off of 160 minutes

velvia:
@chitang ok, now initially_claimed_at is just set on every retry.
@sokrahta defaults bumped up.

coordinator/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
sokrahta: :shipit: let's see how it works and if improvements are needed there are other options!

PR #53, opened 2015-06-03
log config

chitang:


velvia:
:+1:   LGTM but the build failed.

znep:
We only want this for the secondary watcher though, right?  Doesn't data-coordinator get request id and resource passed through?

chitang:
Marc is right.  Will fix it.

PR #55, opened 2015-06-11
instances config for propagateToSecondary()

sokrahta:


znep:
What do you think of instead changing the code to lookup the secondary instances in the database instead of from the config to avoid the stub configs?  There should already be a method that looks up a secondary by name in the db.

sokrahta:
@znep - I saw a way to get instances out of the groups config daf33fc, and then proceeded to query truth c3573df; I think that's correct, but if you like the smaller change I can revert 1 and merge.

sokrahta:
probably need to put this somewhere?
```
insert into secondary_groups_config values ('pg', true, 'pg', 1);
insert into secondary_groups_config values ('spandex', false, 'spandex', 1);
```

sokrahta:
validated on my local stack

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -1,35 +1,34 @@
...
 import org.joda.time.DateTime
+
 import scala.util.Random
 
 class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
   val log = org.slf4j.LoggerFactory.getLogger(classOf[Main])
 
   def ensureInSecondary(storeId: String, datasetId: DatasetId): Unit =
-    for(u <- common.universe) {
+    for { u <- common.universe } {
```
znep: What is wrong with parenthesis here and in other places?  My reading of the scala style guide recommends parenthesis in these cases... http://docs.scala-lang.org/style/control-structures.html#comprehensions

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryInfo.scala
```diff
@@ -1,11 +1,31 @@
...
+      new DateTime(rs.getTimestamp("next_run_time").getTime),
+      rs.getInt("interval_in_seconds"))
+}
+
+case class SecondaryGroupInfo(groupId: String, isDefault: Boolean, instances: Set[String], numReplicas: Int)
+object SecondaryGroupInfo {
+  def apply(rs: ResultSet): SecondaryGroupInfo =
+    SecondaryGroupInfo(rs.getString("id"),
+      rs.getBoolean("is_default"),
+      rs.getString("instances").split("[ ,;]").toSet.map((s: String) => s.trim),
```
znep: Can we actually model this in the database in an array or something?  Feels kind of dirty saying "oh, this is a string using whatever delimiters we want to throw in".

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryInfo.scala
```diff
@@ -1,11 +1,31 @@
...
 
+import java.sql.ResultSet
+
 import org.joda.time.DateTime
 
-case class SecondaryConfigInfo(storeId: String, nextRunTime: DateTime, runIntervalSeconds: Int)
+case class SecondaryInstanceInfo(storeId: String, nextRunTime: DateTime, runIntervalSeconds: Int)
+object SecondaryInstanceInfo {
+  def apply(rs: ResultSet): SecondaryInstanceInfo =
+    SecondaryInstanceInfo(rs.getString("store_id"),
```
znep: Should this sql specific stuff be somewhere in the sql package instead of the base object?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryInfo.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    for {
+      stmt <- managed(conn.prepareStatement(sql))
+      rs <- managed(stmt.executeQuery())
+    } yield accumulate(Set.empty, rs)
+  }
+
+  def instance(storeId: String): Option[SecondaryInstanceInfo] = {
+    val sql = """
+      SELECT store_id, next_run_time, interval_in_seconds
```
znep: nit: Missing `|`s here?

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
velvia: I remember at one point we decided to use braces in all situations....  not a biggie either way of course.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -136,14 +136,13 @@ class Service(serviceConfig: ServiceConfig,
...
-        case "_DEFAULT_" => defaultSecondaryGroups.foreach(ensureInSecondaryGroup(_, datasetId))
-        case groupRe(g) if serviceConfig.secondary.groups.contains(g) => ensureInSecondaryGroup(g, datasetId)
-        case secondary if secondaries(storeId) => ensureInSecondary(secondary, datasetId)
+      for { u <- common.universe } yield { storeId match {
+        case "_DEFAULT_" => u.secondaryInfo.defaultGroups.foreach(g => ensureInSecondaryGroup(g.groupId, datasetId))
+        case groupRe(g) if u.secondaryInfo.groups.exists(_.groupId == g) => ensureInSecondaryGroup(g, datasetId)
+        case secondary if u.secondaryInfo.instance(storeId).isDefined => ensureInSecondary(secondary, datasetId)
         case _ => return NotFound
-      }
+      }}
```
znep: This formatting seems a little funky?

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -126,7 +126,7 @@ class Service(serviceConfig: ServiceConfig,
     override def post = doUpdateVersionInSecondary
 
     def doGetDataVersionInSecondary(req: HttpRequest): HttpResponse = {
-      if(!secondaries(storeId)) return NotFound
+      if (for { u <- common.universe } yield u.secondaryInfo.instance(storeId).isEmpty) return NotFound
```
znep: This check could probably just be removed entirely it seems?

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -59,8 +60,7 @@ class Service(serviceConfig: ServiceConfig,
               datasetContents: (DatasetId, Option[String], CopySelector, Option[UserColumnIdSet],
                                 Option[Long], Option[Long], Precondition, Option[DateTime], Boolean) =>
                                datasetContentsFunc => Exporter.Result[Unit],
-
-              secondaries: Set[String],
+              common: SoQLCommon,
```
znep: Exposing SoQLCommon into Service seems like a bit of a violation of the current (somewhat strange) way that separation of concerns is handled in Service.  In the existing code Service has no notion of a transaction and more or less calls a single function for each that handles the entire database side of things with a single "universe" being created for each call.   

The simplest way to follow that pattern would be to have the function that is passed in be able to do the work that requires looking up the secondaries itself... then it could all be done with the same universe to keep the notion of "one transaction per call".  eg. for doGetSecondaryManifest have datasetsInStore have a way to return an indication that the store doesn't exist.  

That said... the whole way Main and Service interact seems a bit strange to begin with, and in this particular case isn't _actually_ that bad as far as opening multiple transactions, etc. so maybe it isn't worth worrying about.  On the other hand, I wouldn't want someone else to come along and start using common.universe to do more complicated operations such as update operations from within Service.

znep:
Reviewed up to 19f844a

A few general comments as well:

* Can we get rid of SecondaryGroupConfig and anything else that is now obsolete?
* Have you thought through a plan for deploying this to different environments with the right configs?  Unfortunately this is a case that requires "run migration, populate env specific data, then deploy new version" which is messier than most migrations.

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20150615-add-table-secondary-groups-config.xml
```diff
@@ -0,0 +1,25 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="krah" id="20150615-add-column-secondary-group">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="secondary_groups_config"/>
+            </not>
+        </preConditions>
+        <createTable tableName="secondary_groups_config">
+            <column name="group_id" type="VARCHAR(40)">
```
znep: I'm tempted to suggesting calling that group_name throughout instead of id since it isn't really an id, but that is a bit of just a nit perhaps.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
kaidamasaki: I think that parens might be clearer, since it's just a loop.

Though, I really that scala's `for` is overloaded.

sokrahta:
@znep thanks for the comments, all addressed in commits from 89b9bfc to 476c569.
wrt manual config I think we need to get these statements to run once per environment, from the config values in data-coordinator. But is there a better way to transition this data from the config files into the config truth table?

local: 
```sql
insert into secondary_groups_config values ('pg', true, '{pg}', 1);
insert into secondary_groups_config values ('spandex', false, '{spandex}', 1);
```

staging:
```sql
insert into secondary_groups_config values ('pg-aws', true, '{pg1-staging,pg2-staging,pg3-staging}', 2);
insert into secondary_groups_config values ('spandex', false, '{spandex}', 1);
```
rc:
```sql
insert into secondary_groups_config values ('pg-aws', true, '{pg1-rc,pg2-rc,pg3-rc}', 2);
insert into secondary_groups_config values ('spandex', false, '{spandex}', 1);
```
prod:
```sql
insert into secondary_groups_config values ('pg', true, '{sodapg1,sodapg3}', 2);
insert into secondary_groups_config values ('pg-aws-a', true, '{pg3-a-prod,pg5-a-prod,pg7-a-prod}', 1);
insert into secondary_groups_config values ('pg-aws-b', true, '{pg2-b-prod,pg6-b-prod,pg8-b-prod}', 1);
insert into secondary_groups_config values ('spandex3', false, '{spandex3}', 1);
```
eu-west-1-prod:
```sql
insert into secondary_groups_config values ('pg-aws', true, '{pg1-a-eu-west-1-prod,pg2-b-eu-west-1-prod}', 2);
insert into secondary_groups_config values ('spandex', false, '{spandex}', 1);
```

sokrahta:
Maybe the right way to do the config transition is the next deployment of data-coordinator still has secondary groups in config files and on startup looks at the secondary_groups_config table to one-time insert these values. Then a later version will delete the local config.

Local dev stack already has a place for inserting secondary config into the database.
New environment setup also specifies inserting these config into the database.

Your thoughts @znep?

PR #59, opened 2015-07-23
Ability to delay the deletion of a table by x days:

NathalieKaligirwa:
- added default value to reference.conf
- modified Tablecleanup function to accept a finiteduration type.
- Modified calls to SoQLCommon to include a tableCleanupDelay parameter.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/universe/sql/SqlTableCleanup.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/universe/sql/SqlTableCleanup.scala
@@ -2,10 +2,11 @@ package com.socrata.datacoordinator.truth.universe
 package sql
 
 import java.sql.Connection
+import scala.concurrent.duration.FiniteDuration
 
 import com.rojoma.simplearm.util._
 
-class SqlTableCleanup(conn: Connection, daysDelay: Int = 1) extends TableCleanup {
+class SqlTableCleanup(conn: Connection, daysDelay: FiniteDuration) extends TableCleanup {
   val log = org.slf4j.LoggerFactory.getLogger(classOf[SqlTableCleanup])
   def cleanupPendingDrops(): Boolean = {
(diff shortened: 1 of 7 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/universe/sql/SqlTableCleanup.scala
```diff
@@ -2,10 +2,11 @@ package com.socrata.datacoordinator.truth.universe
...
 import java.sql.Connection
+import scala.concurrent.duration.FiniteDuration
 
 import com.rojoma.simplearm.util._
 
-class SqlTableCleanup(conn: Connection, daysDelay: Int = 1) extends TableCleanup {
+class SqlTableCleanup(conn: Connection, daysDelay: FiniteDuration) extends TableCleanup {
   val log = org.slf4j.LoggerFactory.getLogger(classOf[SqlTableCleanup])
   def cleanupPendingDrops(): Boolean = {
     using(conn.createStatement()) { stmt =>
```
urmilan: Does the SELECT query actually end up being valid now that you've changed the type from `Int` to `FiniteDuration`? A good way to test this would be to change your config value to something small like 10 seconds, create and delete a dataset, and see whether data coordinator actually drops the table without error.

chitang: Yes, that probably won't work.  The SQL is:
WHERE queued_at < now() - ('$daysDelay day' :: INTERVAL)
May be the var should be named just delay and change the sql to delay.toSeconds.  Handle values that are smaller than a second.  BTW, this change is commented out by latter commit/release.

urmilan: Yeah, we had to comment it out in master to do a production hotfix yesterday.

PR #62, opened 2015-08-05
Nathaliek/core 5400/add table cleanup delay

NathalieKaligirwa:
@urmilan 
Changes to data-coordinator: 
- New class to handle dropping tables in postgres (SqlTableRemover)
- Modifications in PostgresDatasetMap and PostgresUniverse to handle the new class
- Fixed query to add datasets to pending_table_drops (which is no longer used btw). 
- Remove dataset dropper in data-coordinator


coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -253,52 +259,11 @@ object Main {
...
-                }
-              } catch {
-                case e: Exception =>
-                  log.error("Unexpected error while dropping tables", e)
-              }
-            } while(!finished.await(30, TimeUnit.SECONDS))
-          }
-        }
-
-        val logTableCleanup = new Thread() {
```
urmilan: I don't think you want to delete the logTableCleanup thread. It does something completely different to the table dropper.

urmilan:
I think we've overcomplicated things a little here. It would be simpler to just remove all the delayed table drop code and then modify `SqlDatasetDropper` to delete both t-tables and metadata.

urmilan:
Reviewed up to e2b3ade.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.7.15-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.7.11-SNAPSHOT"
```
urmilan: I don't think you meant to decrement the version?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDatasetDropper.scala
```diff
@@ -9,6 +9,9 @@ import com.rojoma.simplearm.util._
 import java.io.Closeable
 import com.socrata.datacoordinator.truth.DatasetIdInUseByWriterException
 
+//This method was used in a previous implementation of table dropper, where data coordinator put tables first in pending_table_drops
+//Then at a later date, deletes all the datasets in pending_table_drops
+//The To dos refer to that task and are no longer relevant, keeping them around for documentation.
```
urmilan: You can delete all the redundant TODOs. They'll just become a source of confusion at a later date.

urmilan:
Reviewed up to  d91112a - just a couple more minor tweaks before we merge.

urmilan:
Oh, wait. We don't use the `tableCleanupDelay` config value for table dropper anymore, right? Let's get rid of that in the code.

PR #68, opened 2015-10-08
Refs CORE-7133

rozap:
* Raise the command-read-limit to 70m to match soda-fountain
and not die on shapefile imports


catstavi:
Thanks for your explanation!

PR #78, opened 2016-01-14
BUG: EN-1135 location type bug.

chitang:
This is a fall out from the task.  upserting null location value fails because a wrong sub-column index was used.


AEscobarCruz:
lgtm, ship it!

PR #79, opened 2016-01-15
Refs EN-1589

rozap:
* Remove nul char from text reps, as it breaks postgres
* integration tests sold separately


catstavi:
:ship:

PR #86, opened 2016-02-14
Support single row export by row id.

chitang:


coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -116,46 +119,86 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
...
+            // optional row id type must be simple which is represented by a single csv string.
+            rowIdRep.decode(IndexedSeq(rid), IndexedSeq(0))
+                    .map(x => Right(Some(x)))
+                    .getOrElse(Left(Exporter.InvalidRowId))
+          case None => // no customer row identifier.  Use system row identifier.
+            val rowIdRep = common.jsonReps(ctx.copyCtx.datasetInfo)(ctx.copyCtx.pkCol_!.typ)
+            rowIdRep.fromJValue(JString(rid)).map(x => Right(Some(x))).getOrElse(Left(Exporter.InvalidRowId))
         }
       }
+    rowIdPossiblyError match {
```
AEscobarCruz: Could you not return Right or Left above in the Match cases or will it make things overly complicated?

AEscobarCruz:
:shipit:

PR #90, opened 2016-03-15
En 172

rjmac:
Make snapshots a truthstore-only thing.

Secondaries will still be told about them, but they will be told lies: namely that snapshots are immediately deleted.


meganschneider:
:shipit:

PR #91, opened 2016-03-15
EN-158: Fix logging in feedback secondary

aerust:
Report the correct number of row updates


secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackSecondary.scala
```diff
@@ -607,7 +607,7 @@ abstract class FeedbackSecondary[CT,CV, RCI <: RowComputeInfo[CV]] extends Secon
             case 200 =>
               // success! ... well maybe...
               val end = System.nanoTime()
-              log.info("Posted mutation script with {} row updates in {}ms", script.length, (end - start) / 1000000)
+              log.info("Posted mutation script with {} row updates in {}ms", script.length - 2, (end - start) / 1000000)
```
JoeNunnelley: could you add comment explaining the -2 magic number?

aerust: oh... too late, it's pretty obvious in the context, the mutation script starts with two lines that are not the row updates. I can do that, next pull request I make for the secondary

PR #92, opened 2016-03-16
EN-3886: Change /manifest/secondaries-of-dataset

aerust:
Enpoint now includes:
 - version of dataset in truth
 - map from secondary store ids to version
 - feedback secondary stores


aerust:
- Still need to test this
- Need to double check that this endpoint isn't being used by something

rjmac:
lgtm pending the todo list above

aerust:
Okay, tested it locally, and all seems well. Now just waiting on checking sumo again.

aerust:
Checked sumo with query `*"secondaries-of-dataset"*` and nothing came back (other than queries I made).

PR #93, opened 2016-03-22
Handle explicit null on location and phone type in NBE upsert.

chitang:


rjmac:
Anything not decodable is null?

rjmac:
ok

PR #95, opened 2016-03-28
Include secondary watcher claim manager uuid in startup message.

znep:
This is dynamically generated in the docker world and is useful to help track which jobs
are claimed where.


bjornbytes:
:shipit:

PR #96, opened 2016-03-29
EN-4304: Reset isolation level of transactions

aerust:
Actually reset the isolation level of transactions
on check-in to the pool.


aerust:
Still need to test this, and figure out if this is enough to fix the full issue of En-4308.

PR #97, opened 2016-03-30
EN-4304: Reset transaction isolation level

aerust:
Reset the transaction isolation level after commit
and rollback to default level READ COMMITTED.


rjmac:
ok

PR #99, opened 2016-04-04
EN-4323: Fix resync with respect to discarded copies

aerust:
Fix resync for datasets with their latest copy being
a discarded copy. That is, actually set the correct
version (the version in truth) in the secondary manifest.

Also removes dead code path for secondaries that
don't want working copies (we have none).

-- Requires: secondaries to handle these changes
   appropriately, in particular: pg-secondary. --

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
@@ -188,9 +188,6 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
             case ResyncSecondaryException(reason) =>
               logger.info("Incremental update requested full resync: {}", reason)
               resync()
-            case _: InternalResyncForPickySecondary =>
-              logger.info("Resyncing because secondary only wants published copies and we just got a publish event")
-              resync()
           }
         case None =>
           drop()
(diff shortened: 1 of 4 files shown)


aerust:
The "actual" work to do is in pg-secondary

My thought is pg-secondary will make sure it has the state that is wants for discarded copies during Secondary.dropCopy(.).

@znep if you are interested.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -333,28 +296,28 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
             // transaction isolation level is not set to REPEATABLE READ
             case Some(datasetInfo) =>
               val allCopies = r.allCopies(datasetInfo) // guarantied to be ordered by copy number
-            val latest = r.latest(datasetInfo) // this is the newest _living_ copy
+              val latestLiving = r.latest(datasetInfo).copyNumber // this is the newest _living_ copy
+              var latest: Option[metadata.CopyInfo] = None
```
rjmac: `val latest = allCopies.lastOption`?

aerust: No, that is the latest (un)published copy.

aerust:
deleting branch in favor of hotfix branch

PR #101, opened 2016-04-12
Fix deprecation errors

kaidamasaki:


rjmac:
Will likely want to explicitly set the RJ dependency to 3.5 (as `JsonArrayIterator.fromEvents` doesn't exist before then).  Otherwise :+1:

kaidamasaki:
Huh, Emacs appears to have cleaned up a trailing newline.

PR #102, opened 2016-04-12
First pass at hardening the start scripts

kaidamasaki:


bin/run_migrations.sh
```diff
@@ -1,21 +1,18 @@
...
 
-if [ ! -e "$JARFILE" ]; then
-    cd "$BASEDIR" && sbt assembly
-    JARS=( $BASEDIR/coordinator/target/scala-2.10/coordinator-assembly-*.jar )
-    # shellcheck disable=SC2012
-    JARFILE=$(ls -t "${JARS[@]}" | head -n 1)
+JARFILE="$(ls -rt coordinator/target/scala-*/coordinator-assembly-*.jar 2>/dev/null | tail -n 1)"
+if [ -z "$JARFILE" ] || find ./* -newer "$JARFILE" | egrep -q -v '(/target/)|(/bin/)'; then
+    cd "$BASEDIR" || { echo 'Failed to change directories'; exit; }
+    nice -n 19 sbt assembly
```
rjmac: This path never re-sets `JARFILE`, so (especially if it came in because of the `-z`, but also if the version was bumped in a `git pull`) running `java -jar` later will not work.  Same with the other two scripts.

bin/run_migrations.sh
```diff
@@ -1,21 +1,18 @@
...
+CONFIG=${SODA_CONFIG:-$BASEDIR/../docs/onramp/services/soda2.conf} # TODO: Don't depend on soda2.conf.
 
-if [ ! -e "$JARFILE" ]; then
-    cd "$BASEDIR" && sbt assembly
-    JARS=( $BASEDIR/coordinator/target/scala-2.10/coordinator-assembly-*.jar )
-    # shellcheck disable=SC2012
-    JARFILE=$(ls -t "${JARS[@]}" | head -n 1)
+JARFILE="$(ls -rt coordinator/target/scala-*/coordinator-assembly-*.jar 2>/dev/null | tail -n 1)"
+if [ -z "$JARFILE" ] || find ./* -newer "$JARFILE" | egrep -q -v '(/target/)|(/bin/)'; then
+    cd "$BASEDIR" || { echo 'Failed to change directories'; exit; }
```
rjmac: I feel like either the `cd "$BASEDIR"` should either always happen (pretty reasonable) or the script should switch back to the previous directory at the end of this block.

bin/run_migrations.sh
rjmac: The script should also probably exit if `sbt assembly` fails.  Might be a good idea to `set -e` at the top just on general principle.

rjmac:
lgtm!

PR #104, opened 2016-04-13
Add "RowsChangedPreview" message to the log

rjmac:
So that secondaries can take special action on large changes.  I'm not sure I like the `copy_map_table_modifications` thing.  In fact I know I don't like it.  But it's safer than trying to `ALTER TABLE copy_map` at deploy-time.


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
```diff
@@ -53,7 +53,18 @@ trait BasePostgresDatasetMapReader[CT] extends `-impl`.BaseDatasetMapReader[CT]
...
+    """SELECT
+      |  system_id, copy_number, lifecycle_stage :: TEXT, data_version, last_modified, table_modifier
+      |FROM
+      |  copy_map
+      |  LEFT OUTER JOIN copy_map_table_modifiers ON copy_map.system_id = copy_map_table_modifiers.copy_system_id
+      |WHERE
+      |  dataset_system_id = ?
+      |  AND lifecycle_stage <> 'Discarded'
+      |ORDER BY
+      |  copy_number DESC
```
chitang: We just added order by data_version desc in case there are more than one copies.

chitang:
lgtm other than the new order by in latest copy.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
rjmac: good point thanks, I'll change that.

chitang: And lookupQuery too.

PR #108, opened 2016-04-20
Fixes bug-bash issues

aerust:
* ingore `secondary_manifest` column `latest_secondary_lifecycle_stage`
* fixes edge case with `secondary_manifest` column `latest_data_version` value
* adds support for dropping a dataset from a secondary store

Refs EN-4323


coordinator/src/main/scala/com/socrata/datacoordinator/resources/DatasetSecondaryStatusResource.scala
```diff
@@ -42,12 +48,19 @@ case class DatasetSecondaryStatusResource(storeIdOpt: Option[String],
   private def doUpdateVersionInSecondary(req: HttpRequest, storeId: String): HttpResponse = {
     val defaultSecondaryGroups: Set[String] = serviceConfig.secondary.defaultGroups
     val groupRe = "_(.*)_".r
-    storeId match {
-      case "_DEFAULT_" => defaultSecondaryGroups.foreach(ensureInSecondaryGroup(_, datasetId))
+    val found = storeId match {
+      case "_DEFAULT_" => defaultSecondaryGroups.forall(ensureInSecondaryGroup(_, datasetId))
```
rjmac: This doesn't do quite the same thing; `forall` will stop at the first `false`.  Since `ensureInSecondaryGroup` is side-effecting, this is probably not what you want.

You probably want to make this return `true` if at least one of them is `true` (and you don't want to use `exists` for that because it'll stop at the first `true`)

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -59,10 +63,25 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
         datasetId,
         secondaryGroupStr)
 
-      newSecondaries.foreach(ensureInSecondary(_, datasetId))
+      newSecondaries.forall(ensureInSecondary(_, datasetId))
```
rjmac: Ditto re: `foreach`/`forall` having different termination semantics.

rjmac:
Looks ok other than those two things

PR #109, opened 2016-04-20
Add `pending_drop` column to `secondary_manifest`

aerust:
Adds column to support dropping a dataset from
a specific secondary store instance.

Refs EN-4323


rjmac:
:+1:

PR #114, opened 2016-04-25
Update secondary-map to the NEWEST data version, not the LATEST

rjmac:
Because the latest might be a discarded copy while the NEWEST might
be an earlier version that has been updated (e.g., if a working copy
is discarded and the remaining published copy is then updated).


aerust:
lgtm

PR #118, opened 2016-04-28
Don't use the new row's data version when updating by trigger

rjmac:
Use the _highest_ data version.  Because that's correct.


aerust:
👍

PR #119, opened 2016-04-29
EN-3709: fix datetime row ids in NBE

meganschneider:


aerust:
lgtm, I assume we want to get this into the release?

PR #120, opened 2016-05-03
EN-5548: We have a data problem in pg secondary where there are more …

chitang:
…than one published copy.

This will help pg query server to pick up the newer copy if there is conflict.


rjmac:
Hrm, there should "never" be more than one published copy.  There's a bug somewhere that's causing that to happen.  But this is an acceptable workaround, probably.  (The other alternative is to order by `data_version desc` instead, which might be better)

PR #121, opened 2016-05-16
EN-5912: Use Java8 base image for dockerizing secondary-watcher

urmilan:
Jenkins jobs are updated to build the code with Java 8. Build passed in Jenkins when I ran it on this branch.

This is part 1 of 2 of updating secondary watchers to run as Java8. Once this is merged, I'll cut a new release/docker image. The base image for PG and spandex secondary watchers can then be bumped to this new version. FYI: @sokrahta


sokrahta:
pg and spandex secondaries will also need to update jenkins configuration when bumping this version

PR #124, opened 2016-06-06
EN-6724: Configure MaxMetaspace for Java8 images

spaceballone:


urmilan:
:shipit: after checks pass

PR #125, opened 2016-06-10
Bump metaspace for secondary-watcher-pg to 256m

urmilan:


znep:
I'd probably set this in https://github.com/socrata/apps-marathon/blob/master/resources/secondary-watcher-pg.json instead where we set other default "deployed" settings, with the notion the default configs in the docker file should be more constrained for non-prod uses.

PR #129, opened 2016-07-07
EN-7507: Throw resync exc when comp col is addd

aerust:
Actually throw resync exception when a computed
column of the matching strategies is added.

Removes using current cookie as previous after
resync... I don't think it is correct to do so.


rjmac:
:+1: -- I definitely think you're right about not using the old cookie too.

PR #132, opened 2016-07-18
Don't require that instance-names be passed in at all.

rjmac:
We have them in the config file after all.


aerust:
lgtm :boat: Though why does the build fail?

rjmac:
Jsut a transient thing; I started a rebuild and it finishesd fine.

PR #133, opened 2016-07-26
Remove some leftover userspace-backup code

rjmac:
This has caused a deadlock at least once, and it isn't used anywhere anymore.  Most of the code that deals with it has long since been removed.  It's just a stray leftover.


aerust:
lgtm :sheep:

PR #134, opened 2016-07-28
EN-6889: Stop resyncing on computed column creation

aerust:
Export source column rows from core to compute values
for newly created computed columns:
 * Changes retry limit config names
 * Changes the name of cookie field

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Exporter.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Exporter.scala
@@ -20,6 +20,7 @@ object Exporter {
   case object PreconditionFailedBecauseNoMatch extends Result[Nothing]
   case class NotModified(etags: Seq[EntityTag]) extends Result[Nothing]
   case object InvalidRowId extends Result[Nothing]
+  case class UnknownColumns(columns: Set[UserColumnId]) extends Result[Nothing]
   case class Success[T](x: T) extends Result[T]
 
   def export[CT, CV, T](u: Universe[CT, CV] with DatasetReaderProvider,
@@ -48,7 +49,10 @@ object Exporter {
               NotModified(Seq(entityTag))
             case _ =>
               val selectedSchema = columns match {
-                case Some(set) => schema.filter { case (_, ci) => set(ci.userColumnId) }
+                case Some(set) =>
+                  val unknown = set.toSet -- schema.values.map(_.userColumnId)
+                  if (unknown.nonEmpty) return UnknownColumns(unknown)
+                  schema.filter { case (_, ci) => set(ci.userColumnId) || copyCtx.pkCol_!.systemId == ci.systemId }
                 case None => schema
               }
               val properSchema = (copyCtx.userIdCol ++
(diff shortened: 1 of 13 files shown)


aerust:
I still need to handle the 400 from DC case and thread a what should be a 400 through DC that is a 500

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
```diff
@@ -0,0 +1,285 @@
...
+    JObject(Map(
+      "c" -> JString("normal"),
+      "user" -> JString(user)
+    )),
+    JObject(Map(
+      "c" -> JString("row data"),
+      "update_only" -> JBoolean.canonicalTrue,
+      "nonfatal_row_errors" -> JArray(Array(JString("insert_in_update_only")))
+    ))
+  )
```
rjmac: Generally, with modern rojoma-json, it's better to use the json interpolator than building objects by hand.  Much nicer to read, and just as safe (or unsafe..)

coordinator/src/main/scala/com/socrata/datacoordinator/service/Exporter.scala
```diff
@@ -48,7 +49,10 @@ object Exporter {
               NotModified(Seq(entityTag))
             case _ =>
               val selectedSchema = columns match {
-                case Some(set) => schema.filter { case (_, ci) => set(ci.userColumnId) }
+                case Some(set) =>
+                  val unknown = set.toSet -- schema.values.toSet.map { col: ColumnInfo[CT] => col.userColumnId }
```
rjmac: This can probably be `....map(_.userColumnId)`

aerust: Ah yeah

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
```diff
@@ -0,0 +1,346 @@
...
+  // this may throw a ReplayLaterSecondaryException or a BrokenDatasetSecondaryException
+  def feedback(rows: Iterator[Row[CV]],
+               targetColumns: Set[UserColumnId] = cookie.strategyMap.keySet,
+               resync: Boolean = false): FeedbackCookie = {
+    val width = cookie.columnIdMap.size
+    val size = batchSize(width)
+
+    log.info("Feeding back rows with batch_size = {} for target computed columns: {}", size, targetColumns)
+
+    def b(retriesLeft: Int, message: String): Int = {
```
rjmac: "b"?

rjmac: I can see a name like that if it's used many times, but I don't ese that this is used at all...

aerust: It's used twice /shrug

aerust:
Robert says okay, once code review is addressed

PR #138, opened 2016-08-11
Add `unsafeCreateDatasetAllocatingSystemId`

rjmac:
PG secondary needs to use this when initially creating its dataset, so
that it can create the correct initial copy when it is not copy #1.


znep:
🍨  Do we have any changes that will require shipping this off a branch for safety?

rjmac:
I think we might; there's a small amount of alexa-stuff on top of 3.0.50 (which is the current prod version) that touches things which might be visible in DC.

PR #141, opened 2016-08-23
EN-5687: add index on secondary_manifest.went_out_of_sync_at

meganschneider:
Hopefully this will help the initial claim of a dataset work better;
this query is involved in assorted deadlocks. Note that this is not
expected to fix the deadlock issue on its own, but this query being
inefficient isn't helping that problem.


meganschneider:
To be clear: not entirely certain this is the best index to be creating to help. Also not sure how to tell without manually applying this in prod.

meganschneider:
With experimentation in copying the production table, we should definitely NOT create this index, because then postgres does intensely dumb things.

PR #142, opened 2016-08-29
EN-5687: hopefully fix secondary_manifest claim deadlocks

meganschneider:
Change claim query to run on primary key (by using a subselect of original
filters), which will be locking at most one row instead of potentially
hundreds or thousands of rows. If this is the cause of -all- of the
deadlock issues (we see others on attempting to update claimedAt as
well), then yay. If not, then further fixes will be needed.


rjmac:
:+1:

PR #143, opened 2016-09-12
delete invalid rollup on publish

chitang:
validate and drop invalid rollups on publish

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
@@ -6,16 +6,19 @@ import com.rojoma.json.v3.codec.JsonEncode
 import com.socrata.datacoordinator.util.CopyContextResult
 import com.socrata.soql.environment.ColumnName
 import org.joda.time.DateTime
-import com.rojoma.simplearm.{SimpleArm, Managed}
-
+import com.rojoma.simplearm.{Managed, SimpleArm}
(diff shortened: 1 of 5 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
```diff
@@ -372,10 +378,31 @@ object DatasetMutator {
...
+          datasetMap.schema(copyInfo).values.map { colInfo =>
+            (new ColumnName("_" + colInfo.userColumnId.underlying), colInfo.typ)
+          }(collection.breakOut)
+        val prefixedDsContext = new com.socrata.soql.environment.DatasetContext[CT] {
+            val schema: OrderedMap[ColumnName, CT] = OrderedMap(prefixedSchema: _*)
+        }
+
+        datasetMap.rollups(copyInfo).foreach { (ru: RollupInfo) =>
+          try {
+            soqlAnalyzer.analyzeFullQuery(ru.soql)(prefixedDsContext)
```
rjmac: This will re-evaluate `soqlAnalyzer` for each rollup.  Probably not a big deal but just making sure this is deliberate..

rjmac:
Other than that one comment, :+1: from me.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
chitang: I mix up between soql analyzer isn't thread safe but it can be reused.  I will move it outside the loop.

PR #145, opened 2016-10-14
EN-11190: Create index on secondary_manifest for efficient querying by secondary watcher.

znep:
It isn't perfect because we don't quite express things in the right
way in secondary_manifest, but is much improved.  This normally
doesn't make a big difference, but is very very significant if you have
old transactions preventing vacuuming of secondary_manifest combined with a
lot of updates.

### DIFF
--- a/coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20161014-add-secondary-manifest-queue-index.xml
+++ b/coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20161014-add-secondary-manifest-queue-index.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="marcs" id="20161014-add-secondary-manifest-queue-index">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                select count(*) from pg_indexes where tablename='secondary_manifest' and indexname='secondary_manifest_is_a_queue'
+            </sqlCheck>
+        </preConditions>
+        <sql>
+            CREATE INDEX secondary_manifest_is_a_queue ON secondary_manifest (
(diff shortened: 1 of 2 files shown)


znep:
Note this was created in fedramp-prod some weeks ago, this is just backfilling the migration.

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20161014-add-secondary-manifest-queue-index.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="marcs" id="20161014-add-secondary-manifest-queue-index">
+        <preConditions onFail="MARK_RAN">
```
chitang: Does <indexExists> not work?

znep: In what way?  Seemed to work when I ran it...

znep: No, I can't seem to get indexExists to work... it just thinks the index doesn't exist and tries to create it.  There is a similar comment for an older core migration... https://github.com/socrata/core/blob/c872586a038148b2e0e602163a05df494c735b13/unobtainium/src/main/resources/config/migrations/20110311-split-index-and-accumulate.xml#L46-L46

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20161014-add-secondary-manifest-queue-index.xml
```diff
@@ -0,0 +1,24 @@
...
+            <sqlCheck expectedResult="0">
+                select count(*) from pg_indexes where tablename='secondary_manifest' and indexname='secondary_manifest_is_a_queue'
+            </sqlCheck>
+        </preConditions>
+        <sql>
+            CREATE INDEX secondary_manifest_is_a_queue ON secondary_manifest (
```
znep: Can't run `CREATE INDEX CONCURRENTLY` within a transaction unfortunately...

PR #147, opened 2016-10-19
EN-11329: Fix 400s from DC for posting deleted rows

aerust:
Use the `"no_such_row_to_update"` non-fatal error option
in mutation scripts.


NathalieKaligirwa:
lgtm if tests pass :)

PR #154, opened 2017-01-25
EN-13224 Don’t resync in parallel for multiple stores.

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryManifest.scala
```diff
@@ -383,4 +385,30 @@ class SqlSecondaryManifest(conn: Connection) extends SecondaryManifest {
       }
     }
   }
+
+  def lockResync(datasetId: DatasetId, storeId: String, groupName: String): Int = {
+    using(conn.prepareStatement("INSERT INTO resync(dataset_system_id, store_id, group_name) values(?, ?, ?)")) { stmt =>
+      val savepoint = conn.setSavepoint()
```
aerust: So, I didn't give "geocoding" secondary a group name, but I guess that is okay?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -265,6 +265,26 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
+      * Serialize resync in the same store group
+      * because resync will cause reads to be unavailable (at least in pg)
+      */
+    def resyncSerially(): Unit = {
+      val sm = u.secondaryManifest
+      try {
+        // resync immediately calls commit before actual work is done.
+        // That allows us to see resync in progress from the resync table.
+        if (sm.lockResync(datasetId, secondary.storeId, secondary.groupName) > 0) {
+          resync()
```
aerust: so if there is already a lock... what happens exactly? I don't think the the worker will release it's claim properly. I think that needs to be pulled out of resync() and put here.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryManifest.scala
chitang: It will be ok-ish.  Its group default to "" and will be mixed with others which do not set the group.  We probably should add group name unless you have a reason not to.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
chitang: It will throw a ResyncLaterSecondaryException.  It should go back to the queue and retry later.

aerust: It will? Won't the INSERT INTO statement just return 0? and not throw a sql exception?

aerust: I think you need an else with the exception being thrown there?

chitang: insert should always return 1 here if it is successful.  If someone else is in resync, a sql exception will happen.  It is either a duplicated key or timeout (after 3s) depending on whether the other process has committed the resync entry.  If it is a sql exception, it should trigger the special re-enqueue.  If it is something else, it should trigger a generic re-enqueue.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -265,6 +265,25 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
+    def resyncSerially(): Unit = {
+      val sm = u.secondaryManifest
+      try {
+        // resync immediately calls commit before actual work is done.
+        // That allows us to see resync in progress from the resync table.
+        sm.lockResync(datasetId, secondary.storeId, secondary.groupName)
+        resync()
+      } catch {
+        case ex: SQLException =>
+          throw ResyncLaterSecondaryException(ex.getMessage)
```
aerust: Part of me wonders if it would be wise to only catch the expected SQLException here

aerust:
I like it, looks good to me :sheep:

PR #156, opened 2017-01-26
EN-13225:Commit before updating secondary-manifest

aerust:
We want to commit before updating the `latest_secondary_data_version`
to end the current transaction during replay to try to avoid deadlocks
when data-coordinator is also updating the `latest_data_version` for
the same row and is writing to the log tables that the secondary-watcher
was also trying to read from.

Deadlocks at the end of replay cause the secondary-watcher to go into
a resync which causes the dataset to not be available for read.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
@@ -380,6 +380,13 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
     }
 
     def updateSecondaryMap(newLastDataVersion: Long): Unit = {
+      // We want to end the current transaction here. We don't want to be holding share locks on data-tables like log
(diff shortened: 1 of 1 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -380,6 +380,10 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
     }
 
     def updateSecondaryMap(newLastDataVersion: Long): Unit = {
+      // We want to end the current transaction here. We don't want to be holding share locks on data-tables like log
```
chitang: nit, can you add commenting that all db activities up to the first commit should be read only?

chitang:
lgtm (just more comment)

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
aerust: Good point, added that :)

PR #158, opened 2017-02-03
EN-13224 fix regression where secondary_manifest.retry_num is not inc…

chitang:
…remented when there is an error in resync.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryManifest.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryManifest.scala
@@ -13,6 +13,7 @@ import scala.collection.immutable.VectorBuilder
 import com.socrata.datacoordinator.truth.metadata
 import com.socrata.datacoordinator.truth.metadata.DatasetInfo
 import com.socrata.datacoordinator.util.PostgresUniqueViolation
+import org.postgresql.util.PSQLException
 
 import scala.concurrent.duration.FiniteDuration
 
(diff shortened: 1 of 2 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/SqlSecondaryManifest.scala
```diff
@@ -396,9 +397,12 @@ class SqlSecondaryManifest(conn: Connection) extends SecondaryManifest {
...
         stmt.executeUpdate()
       } catch {
+        case PostgresUniqueViolation(_*) =>
+          conn.rollback(savepoint)
+          throw ResyncLaterSecondaryException("wait for another resync")
         case ex: SQLException =>
           conn.rollback(savepoint)
-          throw ex
+          throw ResyncLaterSecondaryException(ex.getMessage)
       }
```
aerust: Are we sure we don't want to catch all exceptions and throw a `ResyncLaterSecondaryException` here?

aerust: Maybe log a error if we get something other than a `SQLException`

chitang: If we don't catch other errors, it will be closer to the original behavior.  Isn't that desirable?

chitang: If the error is not caught here, it will be logged some where.

aerust:
Ship it! :boat:

PR #159, opened 2017-02-15
EN-13662 support for adding more truth instances

chitang:
Parameterize data-coordinator instance so that we specify what instance it is.


aerust:
lgtm :shell:

znep:
Can you update the readme with the new config option?

Also we should perhaps set a default in the dockerfile as a generic reasonable default.

znep:
Thanks!  :squashit: and :shipit:

PR #160, opened 2017-02-15
Add dynamic port mapping support.

chitang:


aerust:
We need this for sure? (bit more of a commit message would be nice)

aerust:
Awesome :ship:

coordinator/src/main/scala/com/socrata/datacoordinator/service/DynamicPortMap.scala
```diff
@@ -0,0 +1,17 @@
+package com.socrata.datacoordinator.service
+
+import com.typesafe.scalalogging.slf4j.Logging
+
+trait DynamicPortMap extends Logging  {
```
rjmac: Don't make a trait that extends "Logging" (actually, just don't use the Logging trait at all); the logger it produces will be named after the dynamic type of the _instance_ (as it uses getClass).

coordinator/src/main/scala/com/socrata/datacoordinator/service/DynamicPortMap.scala
```diff
@@ -1,15 +1,17 @@
 package com.socrata.datacoordinator.service
 
-import com.typesafe.scalalogging.slf4j.Logging
+import org.slf4j.Logger
 
-trait DynamicPortMap extends Logging  {
+trait DynamicPortMap {
+
+  val log: Logger
```
rjmac: This will still use the logger of the dynamic class (well, the class of whatever provides the implementation here)... at least it's more explicit about it, but I'd use a `private val log = getLogger(classOf[DynamicPortMap])` so that the log line comes out as originating in `csds.DynamicPortMap` instead of `csds.Main$`

chitang: ok, i thought you are more concerned about the main class.  I'll just remove logger here.

rjmac:
:boat: :)

PR #163, opened 2017-02-21
EN-12729: Avoid deadlocks on `secondary_manifest`

aerust:
When updating multiple rows on the`secondary_manifest`
in `truth` first obtain locks on the rows in a
deterministic order to avoid deadlocks.

For this: order by `dataset_system_id` then `store_id`.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDatasetDropper.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDatasetDropper.scala
@@ -40,9 +40,18 @@ class SqlDatasetDropper[CT](conn: Connection, writeLockTimeout: Duration, datase
   }
 
   protected def updateSecondaryAndBackupInfo(datasetId: DatasetId, fakeVersion: Long) {
-    using(conn.createStatement()) { stmt =>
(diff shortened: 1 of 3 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDatasetDropper.scala
```diff
@@ -40,9 +40,18 @@ class SqlDatasetDropper[CT](conn: Connection, writeLockTimeout: Duration, datase
...
+    // EN-12729 Note: Ordering by store_id to avoid deadlocks here should not be necessary,
+    // but we are going to do it just in case and for consistency.
+    using(conn.prepareStatement(
+      s"""UPDATE secondary_manifest
+         |SET latest_data_version = ?
+         |WHERE (dataset_system_id, store_id) IN (
+         |  SELECT dataset_system_id, store_id FROM secondary_manifest
+         |  WHERE dataset_system_id = ? ORDER BY store_id FOR UPDATE
+         |)""".stripMargin)) { stmt =>
+      stmt.setObject(1, fakeVersion)
```
znep: nit: The coding convention in the rest of data-coordinator is to use the type specific method (eg. `setLong`) if it exists and only use `setObject` when it doesn't (eg. UUID).

I'm not aware of any good reason in particular for that.

aerust:
Okay, should have hit paths than ran all this sql locally... and nothing exploded

PR #164, opened 2017-03-30
EN-15144: Add column for dataset resource_name

aerust:
We want to start storing dataset resource_name in truth
so the secondary-watcher can include the external dataset
ids in messages in AMQ reporting replication complete.

These replication messages are so we can wait to calculate
column stats after replication is complete.

Also, update migrations script to use sbt instead of first
assembling a new jar (because that is slow).

### DIFF
--- a/bin/run_migrations.sh
+++ b/bin/run_migrations.sh
@@ -3,13 +3,9 @@ set -e
 
 # Run data coordinator migrations
 # run_migrations.sh [migrate/undo/redo] [numchanges]
-REALPATH=$(python -c "import os; print(os.path.realpath('$0'))")
-BINDIR=$(dirname "$REALPATH")
-
 CONFIG="${SODA_CONFIG:-/etc/soda2.conf}" # TODO: Don't depend on soda2.conf.
-JARFILE=$("$BINDIR"/build.sh "$@")
 
 COMMAND=${1:-migrate}
 echo Running datacoordinator.primary.MigrateSchema "$COMMAND" "$2"...
 ARGS=( $COMMAND $2 )
-java -Djava.net.preferIPv4Stack=true -Dconfig.file="$CONFIG" -jar "$JARFILE" com.socrata.datacoordinator.primary.MigrateSchema "${ARGS[@]}"
+sbt -Dconfig.file="$CONFIG" "coordinator/run-main com.socrata.datacoordinator.primary.MigrateSchema ${ARGS[@]}"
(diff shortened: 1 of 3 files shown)


coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20170330-add-column-resource-name.xml
```diff
@@ -0,0 +1,18 @@
...
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="Alexa Rust" id="20161006-add-column-resource-name">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="dataset_map" columnName="resource_name"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="dataset_map">
+            <column name="resource_name" type="VARCHAR(128)">
+                <constraints nullable="true"/>
```
chitang: null = true is default.  But if you still want to be explicit, I am fine.

bin/run_migrations.sh
```diff
@@ -3,13 +3,9 @@ set -e
...
-BINDIR=$(dirname "$REALPATH")
-
 CONFIG="${SODA_CONFIG:-/etc/soda2.conf}" # TODO: Don't depend on soda2.conf.
-JARFILE=$("$BINDIR"/build.sh "$@")
 
 COMMAND=${1:-migrate}
 echo Running datacoordinator.primary.MigrateSchema "$COMMAND" "$2"...
 ARGS=( $COMMAND $2 )
-java -Djava.net.preferIPv4Stack=true -Dconfig.file="$CONFIG" -jar "$JARFILE" com.socrata.datacoordinator.primary.MigrateSchema "${ARGS[@]}"
+sbt -Dconfig.file="$CONFIG" "coordinator/run-main com.socrata.datacoordinator.primary.MigrateSchema ${ARGS[@]}"
```
chitang: We don't want to build before migration?  I don't mind either way.

chitang:
lgtm

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20170330-add-column-resource-name.xml
aerust: I want to be explicit, that is why

bin/run_migrations.sh
aerust: It's slow enough that I think it is pretty annoying

PR #166, opened 2017-03-31
EN-1281: add support for more JDBC timeout configs

jtoberon:
And strongly type them right into some classes.

A param we want to use cancelSignalTimeout requires 9.4.1209 or higher:
https://jdbc.postgresql.org/documentation/94/connect.html

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/DataSourceFromConfig.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/DataSourceFromConfig.scala
@@ -18,9 +18,9 @@ class DataSourceConfig(config: Config, root: String) extends ConfigClass(config,
   val password = getString("password")
   val applicationName = getString("app-name")
   val tcpKeepAlive = optionally(getBoolean("tcp-keep-alive")).getOrElse(false)
-  val loginTimeout = optionally(getInt("login-timeout"))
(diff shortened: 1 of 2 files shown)


jtoberon:
I just want to do the equivalent of java8's ifPresent!

aerust:
If this is related to the soql-servers, why are you doing it in data-coordinator? (not that it shouldn't be done in data-coordinator)

jtoberon:
@aerust because soql-servers use data-coordinator's DataSourceFromConfig class.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/DataSourceFromConfig.scala
```diff
@@ -18,6 +18,9 @@ class DataSourceConfig(config: Config, root: String) extends ConfigClass(config,
   val password = getString("password")
   val applicationName = getString("app-name")
   val tcpKeepAlive = optionally(getBoolean("tcp-keep-alive")).getOrElse(false)
+  val loginTimeout = optionally(getInt("login-timeout"))
```
znep: Use getDuration instead of getInt.

jtoberon: good idea. pushed.

PR #169, opened 2017-04-12
Use camel case in messages instead of underscore.  Historically, that…

chitang:
…’s what we have been doing.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/messaging/Message.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/messaging/Message.scala
@@ -22,7 +22,6 @@ sealed abstract class Message
 // We aren't using this message right now, but I have hopes to use it for
 // webhooks or other things that want to know about when NBE replication
 // is complete. So going to leave this here for now.
-@JsonKeyStrategy(Strategy.Underscore)
 case class StoreReplicationComplete(viewUid: ViewUid,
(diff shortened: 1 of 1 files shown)


catstavi:
There's a couple other classes that use the underscore strategy: `ComputationStrategySpec`, and `SecondariesOfDatasetResult`-- is there a reason those are ok to use the underscore strategy but these aren't?

chitang:
ComputationStrategySpec is SODA2.  Eurybates messages use camel case.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/messaging/Message.scala
```diff
@@ -22,7 +22,7 @@ sealed abstract class Message
 // We aren't using this message right now, but I have hopes to use it for
 // webhooks or other things that want to know about when NBE replication
 // is complete. So going to leave this here for now.
-@JsonKeyStrategy(Strategy.Underscore)
+@JsonKeyStrategy(Strategy.Identity)
 case class StoreReplicationComplete(viewUid: ViewUid,
```
aerust: I might just omit the JsonKeyStrategy, Identity is the default

aerust:
Can you include in the commit message that it is what we do historically for Eurybates just to be clear. Also nice to keep first line of commit messages under 60 characters.

aerust:
looks great :ship:

PR #170, opened 2017-04-19
Explicitly pin Scala version.

znep:
This is a good thing to do anyway, but the specific reason for doing
it is it looks like different versions are fighting resulting in
strange "java.lang.ClassNotFoundException: scala.Some" errors.

I briefly looked at upgrading past 2.10, but we are still blocked by the
sbt-scalabuff plugin, which we would need to either fork and update or
replace.


aerust:
if it builds successfully in jenkins, lgtm :ship:

znep:
retest this please

znep:
Had to bump cloudbees to java 8 to make it succeed.

PR #171, opened 2017-05-05
EN-16084 Minimimal change to bump soql-ref version.

chitang:
We want to do this now so that we will not be blocked if there is an urgent need to bump soil-ref for non join related work.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
@@ -4,7 +4,7 @@ package truth
 import com.rojoma.json.v3.ast.{JNumber, JString, JValue}
 import com.rojoma.json.v3.codec.JsonEncode
 import com.socrata.datacoordinator.util.CopyContextResult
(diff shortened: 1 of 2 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
```diff
@@ -396,7 +396,7 @@ object DatasetMutator {
           val analyzer = soqlAnalyzer
           rollups.foreach { (ru: RollupInfo) =>
             try {
-              analyzer.analyzeFullQuery(ru.soql)(prefixedDsContext)
+              analyzer.analyzeFullQuery(ru.soql)(Map(TableName.PrimaryTable.qualifier -> prefixedDsContext))
             } catch {
```
aerust: huh, so we analyze queries as part of creating rollups?

aerust:
okay

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
chitang: yes, rollup is soql

PR #172, opened 2017-05-10
EN-16021: Prevent race condition in "already claimed" assertion.

znep:
Remove the job from our in memory list first, and then unclaim it in the database.


aerust:
lgtm

PR #173, opened 2017-05-10
EN-15963 Join support

chitang:
. add function to get dataset info by resource name
. add field names to schema

These features will be consumed by soql-postgres-adapter to process joins.


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
```diff
@@ -374,6 +374,25 @@ class PostgresDatasetMapReader[CT](val conn: Connection, tns: TypeNamespace[CT],
     }
   }
 
+  def datasetInfoByResourceNameQuery = "SELECT system_id, next_counter_value, locale_name, obfuscation_key, resource_name FROM dataset_map WHERE resource_name = ?"
```
rjmac: soda-fountain does resource name lookups case-insensitively (it keeps a second, case-folded, resource name column).  We might care about that?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/id/ResourceName.scala
```diff
@@ -0,0 +1,22 @@
+package com.socrata.datacoordinator.id
+
+import com.rojoma.json.v3.codec.{DecodeError, JsonDecode, JsonEncode}
+import com.rojoma.json.v3.ast.{JValue, JString}
+
+class ResourceName(val underlying: String) {
+  override def toString = s"ResourceName($underlying)"
+}
```
rjmac: Other places where we care about user-level identifiers like this (specifically for `ColumnName`s), DC uses the types from soql-reference instead of defining its own.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
chitang: We do not have casefolded version in truth/secondary.

rjmac: You can get one though!  `someResourceName.caseFolded` produces it..

chitang: But the database side does not store it.  Well, we can do schema change if we really want.

rjmac: Ah, right, yeah.  soda-fountain definitely has both a resource_name and a resource_name_casefolded column so it can do that.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/SchemaFinder.scala
```diff
@@ -32,6 +29,13 @@ class SchemaFinder[CT, CV](typeSerializer: CT => TypeName, cache: Cache) extends
     val pk = ctx.pkCol_!.userColumnId
     Schema(hash, columns, pk, ctx.datasetInfo.localeName)
   }
+
+  def getSchemaWithFieldName(ctx: DatasetCopyContext[CT]): SchemaWithFieldName = {
+    val hash = schemaHash(ctx)
```
aerust: So these functions are just being added to the library so soql-server can use them?

chitang: Yes, soql-postgresql-adapter uses dc library.

aerust: Okay

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
aerust: Yeah, when we added resource_name to truth... we weren't thinking about doing lookups with it. Probably makes sense to now materialize the column. Though, it will be a tricky migration, but it worked last time (when we shut down all the secondary-watcher etc)

chitang: It isn't really needed at this point.  We should come back to this when it is needed.

PR #177, opened 2017-06-29
EN-17248: delete columns correctly in Feedback S.

aerust:
Calculate deleted columns correctly in Feedback Secondary
to avoid looping infinitely and running out of memory.

Also add unit tests for deleteColumns function.


aerust:
retest this please

PR #178, opened 2017-06-29
EN-17249: Don't recompute columns on WC creation

aerust:
In the Feedback Secondary don't track CreatedColumnEvents
after WorkingCopyEvents as new computed columns and thus
recompute computed columns whenever working copies are created.


aerust:
retest this please

aerust:
Jenkins build finished successfully... not sure why it didn't report that

PR #179, opened 2017-07-06
EN-16800 setChangeLogLockWaitTime to a value smaller (3s) than sql lo…

chitang:
…ck_timeout (30s) so that migrations where runInTransaction="false" should go through fine.

Also bump soql-refernce just to get the latest version which should not change anything.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/migration/Migration.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/migration/Migration.scala
@@ -1,13 +1,10 @@
 package com.socrata.datacoordinator.truth.migration
 
 import liquibase.Liquibase
-import liquibase.database.jvm.JdbcConnection
-import liquibase.logging.LogFactory
+import liquibase.lockservice.LockService
 import liquibase.resource.ClassLoaderResourceAccessor
-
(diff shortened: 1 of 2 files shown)


znep:
What is this setting doing?  What was it before and how was that causing failure?

chitang:
lockService.setChangeLogLockWaitTime(1000 * 3) tells the process to fail and quit if it cannot lock databasechangeloglock in 3 second.  Without this, the default is 5 minutes which causes the other process that has obtained databasechangeloglock to fail.

znep:
Why is the databasechangeloglock table locked by the other process?  Oh...
perhaps a liquibase bug, leaving an open transaction while waiting...
https://github.com/liquibase/liquibase/blob/2.0.0/liquibase-core/src/main/java/liquibase/lockservice/LockService.java#L102-L112


Oh... Looks like that bug is probably still there in the latest
non-6-year-old version of liquibase if I'm reading the code right... don't
love this workaround but don't have a better one that isn't too much work,
so :shipit:

One thing we should confirm, if you haven't is that it exits with a non
zero exit status in this case so the ship.d/run script properly exits on
failure instead of just proceeding and starting the service.

chitang:
Yep, Process finished with exit code 1

PR #183, opened 2017-09-13
EN-17812: Improve big inserts with row identifiers

rjmac:
By making sure that PG keeps the column stats for the PK column up to
date.


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractRepBasedDataSqlizer.scala
```diff
@@ -161,6 +167,20 @@ abstract class AbstractRepBasedDataSqlizer[CT, CV](val dataTableName: String,
         }
       }
     }
+
+    // WTF postgres WHYYY
```
branweb1: lol

PR #185, opened 2017-09-18
EN-18249: Migrate geocoding secondary cookie

aerust:
Migrate geocoding secondary cookie to reflect changes to cookie
logic from commit 2183e78064db3b916af01e6a1321cb2064d0d3ec in
FeedbackSecondary library. This should be run _before_ using the
new version 0.0.17 of secondary-watcher-geocoding and should be
safe to run at anytime prior.


chitang:
:ship:

PR #186, opened 2017-09-22
EN-18250: Hack secondaries-of-dataset endpoint

aerust:
Hack secondaries-of-dataset endpoint to return the min
of the feedback secondary version and the actual version
of a store for all stores.

This is so the views/4x4/replication.json endpoint can return
false for read replication being up to date if feedback
replication is still behind. This is to avoid customers potentially
running into problems if they have scripts checking the read up to
date value as we migrate all existing region coded datasets to
use the geocoding-secondary.


chitang:
lgtm

PR #192, opened 2017-10-12
Periodically update stats, take two.

rjmac:
I think the problem with the last one is that I didn't take into
account the fact that batch size gets smaller (in terms of rows) as
datasets get wider.  So instead let's (a) keep track of _row counts_
instead of _batches_ and (b) decrease the frequency of ANALYZEs we
issue as datasets get larger, so that we only redo them if the row
count changes by a significant amount (25% of the total row count or
10000 rows, whichever is larger).

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/PostgresRepBasedDataSqlizer.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/PostgresRepBasedDataSqlizer.scala
@@ -55,26 +55,37 @@ class PostgresRepBasedDataSqlizer[CT, CV](tableName: String,
     }
   }
 
-  type PreloadStatistics = Long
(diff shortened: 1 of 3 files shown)


chitang:
:shipit:

znep:
Seems reasonable, although probably still doing a lot more vacuums than it needs to as all we probably care about are "almost no rows do full table scan" vs "oh there are actually rows lets use the index", but that shouldn't hurt tooo much.  

I think there is a good chance that an autovacuum analyze running on the table would block this code until it finishes, but that probably isn't a big deal as long as the autovacuum daemon doesn't do the vacuum and analyze in the same transaction since the vacuum can take a long time.

rjmac:
Cool, thanks.  We can definitely tweak the numbers (I was thinking of dropping it down to 50% rather than 25%, for example) and yeah what I'm _really_ after is "please use this index postgres, I put it there for a reason".

PR #193, opened 2017-10-13
EN-19395: Resurrect PostgresDatasetMapWriterTest 

aerust:
* Get the test compiling and passing again.
 * Bring in scalatest so `sbt it:test` runs successfully.

This test will not run on CloudBees but we will try to get
jenkins-build to run it. Now Requires a test database
datacoordinator_test instead of blist_test (core's test db).

### DIFF
--- a/coordinatorlib/src/it/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
+++ b/coordinatorlib/src/it/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
@@ -57,16 +57,37 @@ class PostgresDatasetMapWriterTest extends FunSuite with MustMatchers with Befor
 
   val resourcName = Some("_abcd-1234")
 
+  val dbName = "datacoordinator_test"
+
+  def withPostgresDb(sql: String): Unit = {
+    using(DriverManager.getConnection("jdbc:postgresql://localhost:5432/postgres", "blist", "blist")) { conn =>
+      conn.setAutoCommit(true)
+      using(conn.createStatement()) { stmt =>
+        stmt.execute(sql)
+      }
+    }
+  }
+
   def populateDatabase(conn: Connection) {
     Migration.migrateDb(conn)
   }
 
   def withDb[T]()(f: (Connection) => T): T = {
-    using(DriverManager.getConnection("jdbc:postgresql://localhost:5432/datacoordinator_test", "blist", "blist")) { conn =>
+    using(DriverManager.getConnection(s"jdbc:postgresql://localhost:5432/$dbName", "blist", "blist")) { conn =>
       conn.setAutoCommit(false)
-      populateDatabase(conn)
       f(conn)
     }
+  }
+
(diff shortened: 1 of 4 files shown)


coordinatorlib/src/test/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
```diff
@@ -1,21 +1,47 @@
...
-import com.socrata.datacoordinator.truth.metadata.CopyPair
-import com.socrata.datacoordinator.truth.migration.Migration
 
+/**
+  * Tests the Postgres implementation of DatasetMapWriter
+  *   _____________________________________
+  *  / Beware! Unlike the rest of our unit \
+  *  | tests which use H2, this test runs  |
+  *  | against a test postgres database    |
+  *  \ blist_test with user blist.         /
```
znep: We should probably just change that to have it use a data_coordinator_test database...

coordinatorlib/src/it/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
```diff
@@ -29,12 +55,14 @@ class PostgresDatasetMapWriterTest extends FunSuite with MustMatchers with Befor
...
 
+  val resourcName = Some("1234-abcc")
+
   def populateDatabase(conn: Connection) {
     Migration.migrateDb(conn)
   }
 
   def withDb[T]()(f: (Connection) => T): T = {
-    using(DriverManager.getConnection("jdbc:postgresql://localhost:5432/blist_test", "blist", "blist")) { conn =>
+    using(DriverManager.getConnection("jdbc:postgresql://localhost:5432/datacoordinator_test", "blist", "blist")) { conn =>
```
chitang: Make it drop and create a new database every run?

coordinatorlib/src/it/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
```diff
@@ -29,12 +55,14 @@ class PostgresDatasetMapWriterTest extends FunSuite with MustMatchers with Befor
   }
   val ZeroID = 0L
 
+  val resourcName = Some("1234-abcc")
```
chitang: Although it is too late and not going to happen.  Could we start a name or identifier with an alphabet and not a number?

coordinatorlib/src/it/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMapWriterTest.scala
aerust: Since postgres doesn't support `create if not exists` for databases it would be a little bit more complicated logic than I really want to write for this... we expect our test postgres databases to exist for all our other tests I can think of.

chitang:
🚢

PR #194, opened 2017-10-17
EN-19395: Get PostgresDatasetMapWriterTest running

aerust:
Get PostgresDatasetMapWriterTest running on Jenkins.
There seems to be a syntax error around encoding=UTF8
when the test is run on Jenkins.


chitang:
🚢

PR #195, opened 2017-10-20
EN-7442: Support computed columns with no sources

aerust:
Support handling computed columns with no source columns.

This is to support geocoded computed columns with only default values
and no source columns, since although they are silly there is no
reason for them to be invalid.


chitang:
lgtm, but your choice.

PR #197, opened 2017-10-26
EN-19241: FB secondary nolonger tracks primary key

aerust:
FeedbackSecondary no longer tracks primary key column of datasets,
but just uses the system id.

Migration to reflect changes in the cookie. This migration will
require down time for the geocoding-secondary (i.e. take the
secondary down, run the migration, then start the secondary on
the new version).

The rollback for the migration is not a perfect rollback since it
sets the primary key as the system id :id regardless if that is
true, but it is a good enough recoverable state and safer to run.


chitang:
:shipit: Maybe you can manually backup to a temp table.

aerust:
I might in production

aerust:
But that will get out of date, and not useful

PR #198, opened 2017-10-27
EN-19241: Fix FeedbackSecondary library

aerust:
Always export the system id column from data-coordinator
since if the dataset has a user primary key column set
data-coordinator will not return the system id column
by default. This is because now FeedbackSecondary
nows does everything by system id and not by primary key.

Fixes bug where system id column is not found in export
from data-coordinator when a user primary key is set
on the dataset. Introduced in commit:
2c58c73e4abb6bc50f0f91bd53da350be8f0f896


chitang:
:shipit:

PR #202, opened 2017-11-05
Avoid tight spin loop if we have errors updating SW claims.

znep:
The immediate motivation for this is that when I'm running secondary
watcher locally and sleep my laptop for a while, when it comes back
the connection pool is closed for some reason and secondary watcher
spins in a tight loop trying to update claims.  Hopefully the root
cause of this will be easier to figure out without the tight spin
loop.

Spinning in this loop is also something we could see in production
if the truth database is unavailable, etc.


znep:
retest this please

znep:
retest this please

socrata-pull-request-updater:
This PR should be reviewed by: @socrata/dataplatform

PR #203, opened 2017-11-20
EN-20157 Change document type to use json (from jsonb)

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/NonLookupableRep.scala
```diff
@@ -0,0 +1,21 @@
+package com.socrata.datacoordinator.common.soql.sqlreps
+
+import java.sql.PreparedStatement
+
+import com.socrata.soql.types.SoQLValue
+
+trait NonLookupableRep {
```
aerust: what enforces this won't be called?

chitang: They will be called when we set a column as custom key.  We don't allow document type to be custom key.  json type in pg doesn't even support eq (jsonb does).

rjmac: That should be done by making SoQLDocument not a PKableRep

chitang:
We are hoping that we will upgrade postgres to 9.4+ until this is really needed.

PR #205, opened 2017-11-27
EN-17014: SecondaryWatcher can help move datasets between stores

aerust:


coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20171122-add-secondary-jobs-tables.xml
```diff
@@ -0,0 +1,69 @@
...
+            <column name="dataset_system_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="store_id" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="job_id" type="UUID">
+                <constraints nullable="false"/>
+            </column>
+            <column name="step" type="SMALLINT">
```
cmlaidlaw: This is an innocent question/learning opportunity for me: what would the drawbacks be for using an ENUM here instead so that the meaning of 'step' is also stored in the database as opposed to just a number?

aerust: step is insteaded to be a step in a collocation operation (or some other hypothetical operation). There can be an arbitrary number of steps in a job so an enum would not work.

aerust:
Note to self: using simplearm.v2 instead of simplearm

PR #206, opened 2017-11-28
EN-16556: Data Coordinator + Secondary Watcher can collocate datasets (WIP)

aerust:


coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20171122-add-secondary-jobs-tables.xml
```diff
@@ -0,0 +1,75 @@
...
+        <createIndex tableName="secondary_add_jobs" indexName="secondary_add_jobs_dataset_system_id_store_id">
+            <column name="dataset_system_id"/>
+            <column name="store_id"/>
+        </createIndex>
+        <createIndex tableName="secondary_add_jobs" indexName="secondary_add_jobs_job_id_step" unique="true">
+            <column name="job_id"/>
+            <column name="step"/>
+        </createIndex>
+
+        <createTable tableName="secondary_delete_jobs">
```
chitang: Have we thought about using one job table for both add and delete?  Or even one job row to represent both add and delete?  That seems easier to support chronological ordering in case there is a need.

aerust: We spent some time talking about both those options. Decided against representing an add and a delete together in a single row because we don't want to have to update. This seemed like the simplest option. When data-coordinator makes a decision about how to move things, it can right a delete and add jobs to respective tables. Then the secondary-watcher depending on what it is doing can look for those jobs and delete them once complete.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
```diff
@@ -293,4 +300,43 @@ class Service(serviceConfig: ServiceConfig,
 
 object Service {
   val numThreads = new AtomicInteger
+
+  def apply(serviceConfig: ServiceConfig,
```
znep: Is there a reason for adding this `apply()` method?  So much copy and paste.

coordinator/src/main/scala/com/socrata/datacoordinator/service/ServiceConfig.scala
```diff
@@ -22,6 +26,7 @@ class ServiceConfig(val config: Config, root: String, hostPort: Int => Int) {
   val commandReadLimit = config.getBytes(k("command-read-limit")).longValue
   val allowDdlOnPublishedCopies = config.getBoolean(k("allow-ddl-on-published-copies"))
   val instance = config.getString(k("instance"))
+  val collocationGroup = Try(config.as[Set[String]](k("collocation_group"))).getOrElse(Set.empty)
```
znep: What is this config used for?  Not a big fan of silently throwing away the config if it can't be parsed.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/CollocationLock.scala
```diff
@@ -0,0 +1,37 @@
...
+  extends CollocationLockException(s"Unexpected exception while $action collocation lock", cause)
+
+trait CollocationLock {
+  def acquire(timeoutMillis: Long): Option[CollocationLockTimeout]
+  def release(): Unit
+}
+
+class CuratedCollocationLock(curator: CuratorFramework) extends CollocationLock {
+  // TODO: what is the right type of curator lock thing to use here?
+  val lock = new InterProcessMutex(curator, "/com.socrata/soda/data-coordinator/collocation-lock")
```
znep: Hardcoding this feels a bit icky ... wonder if using the `curator.namespace` config makes sense.

znep: This seems like the right curator lock, as you need a timeout.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/CollocationLock.scala
```diff
@@ -0,0 +1,37 @@
+package com.socrata.datacoordinator.service.collocation
+
+import java.util.concurrent.TimeUnit
+
+import org.apache.curator.framework.CuratorFramework
+import org.apache.curator.framework.recipes.locks.InterProcessMutex
+
+sealed abstract class CollocationLockException(message: String, cause: Exception) extends Exception(message, cause)
+
+case class CollocationLockTimeout(millis: Long)
```
znep: Does it add any value to wrap the boolean return in a whole case class / option / etc. when there is really no information other than it is acquired or not to be had?

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -0,0 +1,275 @@
...
+        else seen
+      }
+
+      Right(CollocatedDatasetsResult(search(current = datasets, seen = datasets)))
+    } catch {
+      case error: RequestError => Left(error)
+    }
+  }
+
+  override def explainCollocation(secondaryGroup: String, request: CollocationRequest):
```
znep: Will come back to this function in more detail later...

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20171122-add-secondary-move-jobs-table.xml
```diff
@@ -0,0 +1,46 @@
...
+            <column name="from_store_id" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="move_from_store_complete" type="BOOLEAN" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+            <column name="to_store_id" type="VARCHAR(64)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="move_to_store_complete" type="BOOLEAN" defaultValueBoolean="false">
```
znep: Maybe make this a timestamp "completed at" to let you do a bit of tracking by querying the table?

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -0,0 +1,275 @@
...
+              val instances = groupConfig.instances
+              val unavailableInstances = groupConfig.instancesNotAcceptingNewDatasets.getOrElse(Set.empty) // TODO: use in logic
+            val replicationFactor = groupConfig.numReplicas
+
+              request match {
+                case CollocationRequest(collocations) if collocations.isEmpty => Right(Right(CollocationResult.canonicalEmpty))
+                case CollocationRequest(collocations) =>
+                  // TODO: verify collocations (datasets exist and columns exist with matching types)
+
+                  def partitionedNodes[T](edges: Set[Set[T]]): Set[Set[T]] = {
```
znep: These functions should probably be pulled out somewhere with some comments and unit testing and to 
split the generic "this is a graph utility function" code from "this is the real logic of what we are trying to do" so it is obvious at first glance and so they can be reasoned about separately.  This sort of logic seems conducive to unit testing, which would also help make it easier to review by providing examples.

coordinator/src/main/scala/com/socrata/datacoordinator/resources/SecondaryManifestsCollocateResource.scala
```diff
@@ -0,0 +1,50 @@
...
+        case Right(request) if explain => coordinator.collocator.explainCollocation(storeGroup, request)
+        case Right(request) => coordinator.collocator.executeCollocation(storeGroup, request)
+        case Left(decodeError) =>
+          log.warn("Unable to decode request: {}", decodeError.english)
+          return BadRequest // TODO: some kind of error response body
+      }
+
+      collocationResult match {
+        case Right(Right(result)) => OK ~> Json(result)
+        case Right(Left(StoreGroupNotFound(_))) => NotFound
```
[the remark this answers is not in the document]
aerust: logged farther down... but could log again.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Service.scala
aerust: Wanted to be able to create Service as a partially applied function... and then pass the things that need access to zookeeper at the end. but I suppose I can do that all in Main.

PR #211, opened 2018-01-09
EN-17024: Add collocation config into docker

aerust:
Add collocation configuration into docker configs.


chitang:
I prefer using alpha for collocation group in sample-data-coordinator.conf to stop the confusion.  But it is fine if you want to merge as is.  :shipit:

aerust:
I agree, but I would prefer to not add extra steps for people with a standard dev setup to start using solo.

PR #212, opened 2018-01-10
EN-17025: secondary-watcher avoids race conditions

aerust:
Have secondary-watcher take a collocation lock during pending drops
of datasets for move jobs to avoid race conditions during execution
of collocation jobs.


secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -175,29 +178,52 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
+        log.info("Attempting to acquire collocation lock for pending drop of dataset.")
+        if (collocationLock.acquire(collocationLockTimeoutMillis)) {
+          log.info("Acquired collocation lock for pending drop of dataset.")
+
+          val movesToStore = moveJobs.jobsToStore(storeId, datasetId)
+
+          if (movesToStore.nonEmpty) {
+            log.error("Upon pending drop of dataset {} from store {} there are unexepected jobs moving to the store: {}",
+              datasetId.toString, storeId.toString, forLog(movesToStore))
+            throw new Exception("Unexpected jobs moving dataset to store when it is pending drop!")
```
chitang: How is this handled eventually?  drop should trump move?

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -175,29 +178,52 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
+          val movesFromStore = moveJobs.jobsFromStore(storeId, datasetId)
+          log.info("Upon pending drop of dataset {} from store {} completed moves for jobs: {}",
+            datasetId.toString, storeId.toString, forLog(movesFromStore))
+          moveJobs.markJobsFromStoreComplete(storeId, datasetId)
 
-      val movesFromStore = moveJobs.jobsFromStore(storeId, datasetId)
-      log.info("Upon pending drop of dataset {} from store {} completed moves for jobs: {}",
-        datasetId.toString, storeId.toString, forLog(movesFromStore))
-      moveJobs.markJobsFromStoreComplete(storeId, datasetId)
+          collocationLock.release()
```
chitang: place release() call in finally block

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
aerust: This shouldn't happen, assuming I have reasoned about things correctly.... I guess if this does start happening the dataset will eventually end-up broken and we can figure out what is wrong and fix it.

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -175,29 +178,58 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
+        } else {
+          log.error("Failed to acquire collocation lock during pending drop of {}")
+          throw CollocationLockTimeout(collocationLockTimeoutMillis)
+        }
+      } catch {
+        case error: CollocationLockError =>
+          log.error("Unexpected error with acquiring collocation lock during pending drop of dataset!", error)
+          throw error
+      } finally {
+        try {
```
chitang: The try block should start right after "if (acquire" success.  release inside the finally block doesn't need a try block.

chitang:
:shipit:

PR #217, opened 2018-01-22
EN-21356: 400 on datasets w/unrecognized instances

aerust:
Data-coordiantor should 400 on datasets with unrecognized instance
names given to the collocation API instead of 500.


aerust:
retest this please

PR #221, opened 2018-01-26
EN-21820: Fix "complex" collocations

aerust:
Fix "complex" collocations where datasets are already collocated
with other datasets.

### DIFF
--- a/coordinator/src/test/scala/com/socrata/datacoordinator/service/collocation/CoordinatedCollocatorTest.scala
+++ b/coordinator/src/test/scala/com/socrata/datacoordinator/service/collocation/CoordinatedCollocatorTest.scala
@@ -347,8 +347,7 @@ class CoordinatedCollocatorTest extends FunSuite with Matchers with MockFactory 
       result.cost should be (Cost(2))
 
       result.moves.size should be (2)
-      result.moves.map(_.datasetInternalName).toSet.size should be (1)
-      result.moves.head.datasetInternalName should be (bravo2)
+      result.moves.map(_.datasetInternalName).toSet should be (Set(bravo2))
       result.moves.map(_.storeIdTo).size should be (2)
     }
   }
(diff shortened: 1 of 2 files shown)


coordinator/src/test/scala/com/socrata/datacoordinator/service/collocation/CoordinatedCollocatorTest.scala
```diff
@@ -299,6 +299,32 @@ class CoordinatedCollocatorTest extends FunSuite with Matchers with MockFactory
     }
   }
 
+  test("explainCollocation for one default store group and a collocated dataset and a non-collocated dataset should move the non-collocated dataset twice") {
```
PiSoup: It would be nice to have a simple illustration of where the dataset will be moving. Just something like:
`bravo2: store5A -> store1A`
or whatever.

coordinator/src/test/scala/com/socrata/datacoordinator/service/collocation/CoordinatedCollocatorTest.scala
```diff
@@ -299,6 +322,67 @@ class CoordinatedCollocatorTest extends FunSuite with Matchers with MockFactory
...
+      expectCDOIAlpha1WithCollocationsComplex(coordinator) // group for alpha.1
+      expectCDOIDatasetsWithNoCollocations(coordinator, Set(bravo2)) // group for bravo.2
+    }) { case (collocator, _) =>
+      val result = collocator.explainCollocation(storeGroupA, request(Seq((alpha1, bravo2)))).right.get
+      result.id should be (None)
+      result.status should be (Approved)
+      result.message should be (Approved.message)
+      result.cost should be (Cost(2))
+
+      result.moves.size should be (2)
```
PiSoup: It might be worth verifying the specific moves instead of the general shape. Currently something moving to a bogus `storeIdTo` would pass.

aerust: Yeah, problem is I have some random generators... I can fix it though

aerust: So I don't know exactly what the moves may be, but I can check that the destination stores are in the correct subset

PR #225, opened 2018-01-31
EN-21910: Pass JobId for Collocation

PiSoup:
Take collocation job id as a passed parameter rather than generating it in data-coordinator. This allows core to more easily maintain a manifest of collocations.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationSodaResource.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationSodaResource.scala
@@ -48,17 +48,10 @@ abstract class CollocationSodaResource extends SodaResource {
   def datasetNotFound(datasetInternalName: DatasetInternalName, resp: HttpResponse = NotFound): HttpResponse =
     errorResponse(resp, CollocationError.DATASET_DOES_NOT_EXIST, "dataset" -> JString(datasetInternalName.underlying))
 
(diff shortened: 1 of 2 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationSodaResource.scala
```diff
@@ -47,23 +48,34 @@ abstract class CollocationSodaResource extends SodaResource {
   def datasetNotFound(datasetInternalName: DatasetInternalName, resp: HttpResponse = NotFound): HttpResponse =
     errorResponse(resp, CollocationError.DATASET_DOES_NOT_EXIST, "dataset" -> JString(datasetInternalName.underlying))
 
-  def withBooleanParam(name: String, req: HttpRequest)(handleRequest: Boolean => HttpResponse): HttpResponse = {
-    val param = Option(req.servletRequest.getParameter(name)).getOrElse("false")
+  def withTypedParam[T](name: String, req: HttpRequest, defaultValue: T)(handleRequest: T => HttpResponse): HttpResponse = {
+    val param = Option(req.servletRequest.getParameter(name)).getOrElse(defaultValue.toString)
     try {
-      handleRequest(param.toBoolean)
+      val parsedParam = defaultValue match {
```
aerust: might be nice to have this function instead take a function from `String => T` so it can be generic without this match case.

PR #230, opened 2018-02-07
EN-21686: Add secondary-metrics API to Coordinator

aerust:
Add secondary-manifest/metrics endpoint to Coordinator service
to be used in explain and initiate of collocation by the Collocator
service.

Add the dataset size from the secondary-metrics API to Cost in
collocation, but don't do anything with it yet.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
@@ -26,5 +26,7 @@ object Cost {
 
   val Zero: Cost = Cost(moves = 0, totalSizeBytes = 0L, moveSizeMaxBytes = Some(0L))
 
-  val Unknown: Cost = Cost(moves = 1, totalSizeBytes = Long.MinValue)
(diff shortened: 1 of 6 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
```diff
@@ -3,18 +3,28 @@ package com.socrata.datacoordinator.service.collocation
...
 
 object Cost {
   implicit val encode = AutomaticJsonEncodeBuilder[Cost]
 
-  val Zero: Cost = Cost(moves = 0)
+  def max(x: Cost, y: Cost): Cost = Ordering[Cost].max(x, y)
+
+  val Zero: Cost = Cost(moves = 0, totalSizeBytes = 0L, moveSizeMaxBytes = Some(0L))
+
+  val Unknown: Cost = Cost(moves = 1, totalSizeBytes = Long.MinValue)
```
znep: What is the intended use of Unknown?  Would totalSizeBytes=0 make more sense?   Seems a bit strange to say "we don't know the size so we will assume it is impossibly small and have a strong bias towards any move that involves moving it".

PR #233, opened 2018-02-21
EN-21686: Add size to cost vector in collocation

aerust:
Add total size and maximun (individual) move size to
cost vector for collocation in addition to the number
of moves. The ordering for cost is configurable with
weights.

For the pg-secondary implementation, the size here is
the total size of t-tables with indices of datasets
on the secondary-stores.

Includes what I think is a reasonable starting point
for weights in `reference.conf`.


coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -579,6 +579,12 @@ object Main extends DynamicPortMap {
         val secondaryManifestsResource = SecondaryManifestsResource(_: Option[String], secondaries,
           operations.datasetsInStore, common.internalNameFromDatasetId)
 
+        implicit val weightedCostOrdering: Ordering[Cost] = WeightedCostOrdering(
```
PiSoup: I'm not sure it's worth pulling the cost ordering function out of the collocator. The weights in the config already give some levers to adjust so this seems like overkill. 
I think that it might be more valuable for code readability to keep the business logic of the collocation tightly coupled.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -114,24 +115,20 @@ class CoordinatedCollocator(collocationGroup: Set[String],
     storesInGroup
   }
 
-  protected def movesFor(group: Set[DatasetInternalName], storesFrom: Set[String], storesTo: Set[String]): Seq[Move] = {
+  protected def movesFor(group: Set[DatasetInternalName],
+                         storesFrom: Set[String],
+                         storesTo: Set[String],
+                         costMap: Map[DatasetInternalName, Cost]): Seq[Move] = {
     assert(storesFrom.size == storesTo.size)
```
PiSoup: Might be worth wrapping this in a handled exception.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -114,24 +115,20 @@ class CoordinatedCollocator(collocationGroup: Set[String],
...
 
     scala.util.Random.shuffle((storesFrom -- storesTo).toSeq).
       zip((storesTo -- storesFrom).toSeq).
       flatMap { stores =>
         val (storeFrom, storeTo) = stores
```
PiSoup: Just extract this on line 126, make it a bit cleaner.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
```diff
@@ -3,11 +3,7 @@ package com.socrata.datacoordinator.service.collocation
 import com.rojoma.json.v3.util.{AutomaticJsonEncodeBuilder, JsonKeyStrategy, Strategy}
 
 @JsonKeyStrategy(Strategy.Underscore)
-case class Cost(moves: Int, totalSizeBytes: Long, moveSizeMaxBytes: Option[Long] = None) extends Ordered[Cost] {
-  override def compare(that: Cost): Int = {
-    this.moves - that.moves
-  }
-
+case class Cost(moves: Int, totalSizeBytes: Long, moveSizeMaxBytes: Option[Long] = None) {
```
PiSoup: It might make more sense to call this something like CostVector since the actual cost is the output of the weight polynomial. As is it's a bit tricky to tell when you're referring to the computed cost vs. the feature vector.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
```diff
@@ -22,11 +18,41 @@ case class Cost(moves: Int, totalSizeBytes: Long, moveSizeMaxBytes: Option[Long]
...
+//
+// movesWeight > totalSizeBytesWeight + moveSizeMaxBytesWeight
+// and movesWeight > 0.0
+case class WeightedCostOrdering(movesWeight: Double,
+                                totalSizeBytesWeight: Double,
+                                moveSizeMaxBytesWeight: Double) extends Ordering[Cost] {
+
+  if (movesWeight <= 0.0) throw new IllegalArgumentException("movesWeight must be positive")
+
+  private def parameterNonNegative(name: String, value: Double): Unit =
```
PiSoup: This is probably overkill since you're not really stopping much code reuse.
`if (totalSizeBytesWeight < 0.0) throw new IllegalArgumentException("totalSizeBytesWeight must be non-negative")` reads much easier.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
```diff
@@ -22,11 +18,41 @@ case class Cost(moves: Int, totalSizeBytes: Long, moveSizeMaxBytes: Option[Long]
...
+    if (value < 0.0) throw new IllegalArgumentException(s"$name must be non-negative")
+
+  parameterNonNegative("totalSizeBytesWeight", totalSizeBytesWeight)
+  parameterNonNegative("moveSizeMaxBytesWeight", moveSizeMaxBytesWeight)
+
+  private val byteWeightSum = totalSizeBytesWeight + moveSizeMaxBytesWeight
+  if (movesWeight <= byteWeightSum)
+    throw new IllegalArgumentException("movesWeight > totalSizeBytesWeight + moveSizeMaxBytesWeight, " +
+      s"but $movesWeight <= $totalSizeBytesWeight + $moveSizeMaxBytesWeight = $byteWeightSum")
```
PiSoup: Probably remove the `= $byteWeightSum` from the message.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Metric.scala
```diff
@@ -20,7 +21,7 @@ case class CoordinatedMetric(collocationGroup: Set[String], coordinator: Coordin
         coordinator.secondaryMetrics(storeId, dataset).fold(throw _, identity).map { metric =>
           Cost(moves = 1, totalSizeBytes = metric.totalSizeBytes)
         }
-      }.reduceOption(Cost.max).getOrElse(Cost.Unknown) // TODO: note we way want to replace Unknown here in the future with and error
+      }.reduceOption(costOrdering.max).getOrElse(Cost.Unknown) // TODO: note we way want to replace Unknown here in the future with and error
```
PiSoup: Why not have it error now?

aerust: would rather have something that works for people trying to use it

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
aerust: I guess the main reason I was doing this is that I also use the ordering in Metric, and I didn't really want to pass the config values in to create the ordering in multiple places.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Cost.scala
aerust: I guess the way I was thinking about it, is that the "cost" is the vector of different values (which actually have some meaning), and the cost is ordered. The computed weighted value is just an implementation detail of an ordering (since its value really doesn't mean anything).

aerust: Maybe I will change my commit message to better reflect that.

PiSoup: Maybe throw a comment where it responds. The reason I was confused is I've been looking at cost from that direction where it is just a number.

PR #234, opened 2018-02-22
EN-21891: Delete entries from secondary_move_jobs

aerust:
Delete entries from secondary_move_jobs when datasets are deleted
since secondary_move_jobs has a foreign key contraint to dataset_map.


chitang:
:shipit:

PR #235, opened 2018-02-22
EN-22550: Drop entries from collocation manifests

aerust:
Drop entries from collocation manifests as datasets are deleted.

New route:

 - `DELETE /collocation-manifest/{instance}/{dataset_system_id}`:
   Drops entries for the given dataset from the collocation manifest
   of the given data-coordinator instance.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationManifestsResource.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationManifestsResource.scala
@@ -1,8 +1,10 @@
 package com.socrata.datacoordinator.resources.collocation
 
+import com.rojoma.json.v3.ast.JString
 import com.rojoma.json.v3.util.{AutomaticJsonCodecBuilder, JsonKeyStrategy, Strategy}
+import com.socrata.datacoordinator.external.CollocationError
 import com.socrata.datacoordinator.id.DatasetInternalName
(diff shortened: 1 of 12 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/CollocationManifestsResource.scala
```diff
@@ -41,4 +46,21 @@ case class CollocationManifestsResource(instanceId: Option[String],
...
+    (instanceId, datasetId.flatMap(DatasetInternalName(_))) match {
+      case (Some(instance), Some(dataset)) =>
+        coordinator.dropCollocationsOnInstance(instance, dataset) match {
+          case None => OK
+          case Some(InstanceNotFound(_)) => instanceNotFound(instance)
+          case Some(_) => InternalServerError
+        }
+      case (None, _) => instanceNotFound("")
+      case (_, None) => errorResponse(
+        BadRequest,
```
chitang: Isn't internal name not found an internal error?

aerust: Not to data-coordinator? It would be an internal error to soda-fountain I think.

chitang: ok

chitang:
lgtm

PR #236, opened 2018-02-22
EN-17026: Collocation API can reject collocation

aerust:
Add limits to collocation cost passed in the collocation
request such that data-coordinator can reject collocation
when the cost exceeds those limits.

Add a new `"limits"` field to the collocation request body
pointing to a Cost object.

Note: this is a breaking change! Since the new limits field
is required.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
@@ -6,14 +6,19 @@ import com.rojoma.json.v3.util._
 import com.socrata.datacoordinator.common.collocation.{CollocationLock, CollocationLockError, CollocationLockTimeout}
 import com.socrata.datacoordinator.id.DatasetInternalName
 import com.socrata.datacoordinator.resources.collocation.{CollocatedDatasetsResult, DatasetNotInStore, SecondaryMoveJobRequest, StoreNotAcceptingDatasets}
+import com.socrata.datacoordinator.service.collocation.CollocationRequest.CostLimits
 import com.socrata.datacoordinator.service.collocation.secondary.stores.SecondaryStoreSelector
 
 import scala.annotation.tailrec
 
 @JsonKeyStrategy(Strategy.Underscore)
-case class CollocationRequest(collocations: Seq[(DatasetInternalName, DatasetInternalName)])
+case class CollocationRequest(collocations: Seq[(DatasetInternalName, DatasetInternalName)],
+                              limits: CostLimits)
 
 object CollocationRequest {
+  type CostLimits = Cost
+
+  implicit val costLimitsDecode = AutomaticJsonDecodeBuilder[CostLimits]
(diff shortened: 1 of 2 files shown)


coordinator/src/test/scala/com/socrata/datacoordinator/service/collocation/CoordinatedCollocatorTest.scala
```diff
@@ -628,7 +647,72 @@ class CoordinatedCollocatorTest extends FunSuite with Matchers with MockFactory
...
+
+      explainShouldBeRejected(expectedRejectionMessage),
+
+      executeShouldBeRejected(expectedRejectionMessage)
+    )
+  }
+
+  testExpectRejectedExplainAndExecuteCollocation(
+    "for collocation with moves exceeding the supplied move limit should reject the collocation",
+     costLimits.copy(moves = 3),
```
PiSoup: It'd be a bit easier to read this if you just created a new costLimit instead of a copy with one param changed. Same for the 2 below.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -6,14 +6,19 @@ import com.rojoma.json.v3.util._
 import com.socrata.datacoordinator.common.collocation.{CollocationLock, CollocationLockError, CollocationLockTimeout}
 import com.socrata.datacoordinator.id.DatasetInternalName
 import com.socrata.datacoordinator.resources.collocation.{CollocatedDatasetsResult, DatasetNotInStore, SecondaryMoveJobRequest, StoreNotAcceptingDatasets}
+import com.socrata.datacoordinator.service.collocation.CollocationRequest.CostLimits
```
PiSoup: Instead of importing this here I would just fully qualify it on line 16. Everything else should already have access to the companion object.

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -6,14 +6,13 @@ import com.rojoma.json.v3.util._
...
 import com.socrata.datacoordinator.resources.collocation.{CollocatedDatasetsResult, DatasetNotInStore, SecondaryMoveJobRequest, StoreNotAcceptingDatasets}
-import com.socrata.datacoordinator.service.collocation.CollocationRequest.CostLimits
 import com.socrata.datacoordinator.service.collocation.secondary.stores.SecondaryStoreSelector
 
 import scala.annotation.tailrec
 
 @JsonKeyStrategy(Strategy.Underscore)
 case class CollocationRequest(collocations: Seq[(DatasetInternalName, DatasetInternalName)],
-                              limits: CostLimits)
+                              limits: com.socrata.datacoordinator.service.collocation.CollocationRequest.CostLimits)
```
PiSoup: Guess I meant partially qualify. `CollocationRequest.CostLimits` should work.

PR #237, opened 2018-02-23
EN-22550: Drop entries from collocation manifests

aerust:
Drop entries from collocation manifests as datasets are deleted.

New route:

 - `DELETE /collocation-manifest/{instance}/{dataset_system_id}`:
   Drops entries for the given dataset from the collocation manifest
   of the given data-coordinator instance.


aerust:
Chi already approved this, I just merged it wrong... 

https://github.com/socrata-platform/data-coordinator/pull/235

PR #238, opened 2018-02-26
EN-17026: Collocation API will not overfill stores

aerust:
Collocation API will not overfill secondary stores.

Reworks secondary store config to include values for the
store capacity. This capacity should be the rough capacity
the store has for row data and indices of datasets, it should
not include capacity needed to metadata.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
@@ -153,7 +153,7 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
         for (u <- common.universe) yield {
           u.datasetMapReader.datasetInfo(datasetId) match {
             case Some(_) =>
-              val moves = u.secondaryMoveJobs.jobs(datasetId).filter { move => groupConfig.instances(move.toStoreId) }
+              val moves = u.secondaryMoveJobs.jobs(datasetId).filter { move => groupConfig.instances.keySet(move.toStoreId) }
 
               Right(SecondaryMoveJobsResult(moves))
             case None =>
@@ -176,17 +176,17 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
               val fromStoreId = request.fromStoreId
               val toStoreId = request.toStoreId
 
-              if (!groupConfig.instances(fromStoreId)) return Left(StoreNotFound(fromStoreId))
-              if (!groupConfig.instances(toStoreId)) return Left(StoreNotFound(toStoreId))
+              if (!groupConfig.instances.keySet(fromStoreId)) return Left(StoreNotFound(fromStoreId))
+              if (!groupConfig.instances.keySet(toStoreId)) return Left(StoreNotFound(toStoreId))
 
-              if (groupConfig.instancesNotAcceptingNewDatasets.getOrElse(Set.empty)(toStoreId)) {
+              if (!groupConfig.instances(toStoreId).acceptingNewDatasets) {
                 log.warn("Cannot move dataset to store {} not accepting new datasets", toStoreId)
                 return Right(Left(StoreNotAcceptingDatasets))
               }
 
               val currentStores =
                 secondariesOfDataset(datasetId).map(_.secondaries.keySet).getOrElse(Set.empty).
-                  filter(groupConfig.instances(_))
+                  filter(groupConfig.instances.keySet(_))
 
               val existingJobs = u.secondaryMoveJobs.jobs(datasetId)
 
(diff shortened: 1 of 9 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/config/SecondaryConfig.scala
```diff
@@ -8,10 +8,11 @@ import net.ceedubs.ficus.FicusConfig._
 // reflection-based horror that doesn't even provide full paths to
 // errors.
 
+case class StoreConfig(storeCapacityMB: Long, acceptingNewDatasets: Boolean)
+
 case class SecondaryGroupConfig(
      numReplicas: Int,
-     instances: Set[String],
-     instancesNotAcceptingNewDatasets: Option[Set[String]]
+     instances: Map[String, StoreConfig]
```
aerust: @znep You may have feelings about this config change.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -449,13 +449,10 @@ object Main extends DynamicPortMap {
...
-    val secondaries: Set[String] = serviceConfig.secondary.groups.flatMap(_._2.instances).toSet
+    val secondaries: Set[String] = serviceConfig.secondary.groups.flatMap(_._2.instances.keySet).toSet
+    // TODO: remove this
     val secondariesNotAcceptingNewDatasets: Set[String] =
-      serviceConfig.secondary.groups.flatMap(_._2.instancesNotAcceptingNewDatasets).flatten.toSet
-    serviceConfig.secondary.groups.foreach { case (name, group) =>
-      if (!group.instancesNotAcceptingNewDatasets.getOrElse(Set.empty).subsetOf(group.instances))
-        throw new Exception(s"For secondary group $name instancesNotAcceptingNewDatasets must be a subset of instances!")
-    }
+      serviceConfig.secondary.groups.flatMap(_._2.instances.filter(!_._2.acceptingNewDatasets).keySet).toSet
```
PiSoup: The anonymous underscore -> tuple makes this pretty tough to parse. Probably worth unwrapping the tuples.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/config/SecondaryConfig.scala
znep: bitmoji

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -235,8 +238,17 @@ class CoordinatedCollocator(collocationGroup: Set[String],
                   Rejected(s"the total size in bytes exceeds the limit ${costLimits.totalSizeBytes}")
                 else if (totalCost.getMoveSizeMaxBytes > costLimits.getMoveSizeMaxBytes)
                   Rejected(s"the max move size in bytes exceeds the limit ${costLimits.getMoveSizeMaxBytes}")
-                else
-                  Approved
+                else {
+                  totalMoves.groupBy(_.storeIdTo).flatMap { case (storeId, moves) =>
+                    val bytesToMove = Move.totalCost(moves).totalSizeBytes
+                    val storeTotalBytes = storeMetricsMap(storeId).totalSizeBytes
+                    val storeCapacityBytes = groupConfig.instances(storeId).storeCapacityMB * 1000 * 1000
```
PiSoup: This might actually need to be `* 1024 * 1024`, not sure what environment these values are going to be coming from. Might also be worth putting a bit of fudge in there anyway. Instead of filling exactly to capacity only allow collocation to bring it within something like 5% of capacity?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/config/SecondaryConfig.scala
aerust: What does that mean?

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
aerust: I was thinking capacity would already have a margin on it, so no need to add extra logic for that here.

aerust:
Needed apps-marathon changes: https://github.com/socrata/apps-marathon/pull/770

coordinator/src/main/scala/com/socrata/datacoordinator/service/collocation/Collocator.scala
```diff
@@ -250,6 +250,12 @@ class CoordinatedCollocator(collocationGroup: Set[String],
                   }.headOption.getOrElse(Approved)
                 }
 
+              totalStatus match {
+                case Rejected(reason) => log.info("Rejecting collocation request because {}", reason)
+                case Approved => log.info("Approving collocation request")
```
PiSoup: Since you're adding logging for this could you also get some info about what is getting collocated

PR #241, opened 2018-03-07
Add solo configuration for phonetic dc cluster

aerust:
Add solo config and start scripts to run the phonetic
truth cluster "alpha", "bravo", and "charlie" for
local development with multiple truth shards.


znep:
Looks ok, but you might want to consider just having `start_dc.sh` and `start_dc_docker.sh` take an optional argument.

aerust:
@znep I totally agree, but then I would have to add functionality to solo for that. (which I might do at some point... but for now i would like to just get this merged.)

znep:
Blah, that is silly but ok.

PR #253, opened 2018-10-10
EN-27755 Support add/drop computation strategy on existing columns

chitang:


rjmac:
I'm not fond of this; the contents of a column with a computation strategy are meaningless without the associated strategy.  It's better to do schema modifications than to mess with the computation strategy.

chitang:
Isn't the computation strategy complete?  What is the missing associated strategy?

znep:
I think the issue is they are unlikely to move all at once?
________________________________

chitang:
A backfill looks really complicated especially if we don't want to re-geocode with an incompatible cache.

chitang:
The complexity is additive and basically boilplate, not multiplicative.

chitang:
probably can restrict this for migration only and revert once migration is over.

rjmac:
(fwiw the code itself looks good; convince me that the approach as a whole works well and I'll happily :+1: this)

PR #255, opened 2018-10-19
Eddie/soql upgrade

EddieCarlsonSocrata:
use soql-reference 2.10.6

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "3.4.31-SNAPSHOT"
+version in ThisBuild := "3.4.31"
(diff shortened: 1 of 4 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "3.4.32-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "3.4.32"
```
PiSoup: Does this need to be manually ticked or will publication take care of it?

EddieCarlsonSocrata:
out of memory error on the jenkins test?

PR #265, opened 2019-03-13
Make it more flexible in controlling replay delay

chitang:


secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -107,12 +107,19 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
           // Prefer cookie handed-back by the exception, but since we write notes in out cookies currently that we don't
           // want to lose, let's avoid overriding those if we are not given Some other cookie string
           manifest(u).markSecondaryDatasetBroken(job, cookie.orElse(job.initialCookie))
-        case rlse@ReplayLaterSecondaryException(reason, cookie) =>
-          if (job.replayNum < maxReplays) {
-            val replayAfter = Math.min(replayWait.toSeconds * Math.log(job.replayNum + 2), maxReplayWait.toSeconds)
+        case rlse@ReplayLaterSecondaryException(reason, cookie, delay) =>
+          val (replayDelay, replayNum) =
+            delay match {
+              case Some(d) => (d, if (job.replayNum >= maxReplays) 0 else job.replayNum)
```
PiSoup: Does this mean that if a delay is set it will never stop attempting to replay? Is that desired?

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -107,12 +107,19 @@ class SecondaryWatcher[CT, CV](universe: => Managed[SecondaryWatcher.UniverseTyp
...
+              case None => (replayWait, job.replayNum)
+            }
+          if (replayNum < maxReplays) {
+            val replayAfter =
+              if (delay.isDefined) replayDelay.toSeconds
```
PiSoup: Is this condition needed? It seems that using else condition in either case would be a bit cleaner.

PiSoup:
How would a user end up setting the replay delay?

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
chitang: yes, the control is in the hand of the secondary plugin if that parameter is used.  We can put a bound in the hand of secondary-watcher too.  It just becomes complicated.  There is no rush to do/solve this problem.  The current way to disable read/write works.  It is just a little cumbersome.

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
chitang: The logarithmic function is for the standard delay value dictate by secondary watcher.  I think if the delay is specified by the plugin, it should be used as is.  Again, this is a debateble issue.  Since we are not merging this to master, we probably don't need a verdict now.

chitang:
I was thinking about adding the field - dataset_internal_name_map.disabled_write in pg secondary.  It already has dataset_internal_name_map.disabled (which is implicitly for read)

PR #267, opened 2019-03-21
EN-31249 Keep secondary_metrics history in truth.  Updated by seconda…

chitang:
…ry watcher.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/PostgresSecondaryMetrics.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/PostgresSecondaryMetrics.scala
@@ -1,12 +1,14 @@
 package com.socrata.datacoordinator.secondary.sql
 
 import java.sql.Connection
+import java.util.{Calendar, Date}
 
 import com.rojoma.simplearm.v2.using
 import com.socrata.datacoordinator.id.DatasetId
 import com.socrata.datacoordinator.id.sql._
 import com.socrata.datacoordinator.secondary.SecondaryMetric
 import com.socrata.datacoordinator.util.PostgresUniqueViolation
+import org.apache.commons.lang.time.DateUtils
 
 class PostgresSecondaryMetrics(conn: Connection) extends SqlSecondaryMetrics(conn) {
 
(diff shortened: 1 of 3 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/PostgresSecondaryMetrics.scala
```diff
@@ -44,5 +46,55 @@ class PostgresSecondaryMetrics(conn: Connection) extends SqlSecondaryMetrics(con
...
+          stmt.execute()
+        }
+      } catch {
+        case uv@PostgresUniqueViolation(_*) =>
+            // there should only be one thread trying to due this at a time since this is only called in the SecondaryWatcher
+            log.error("Inserting into secondary_metrics_history failed after updating no rows", uv)
+      }
+    }
+
+// TODO: replace this with using Postgres UPSERT functionality introduced in 9.5
```
EddieCarlsonSocrata: do you want to keep this whole commented-out section?

EddieCarlsonSocrata:
looks good to me

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/PostgresSecondaryMetrics.scala
chitang: Yes, I already did this for pg 9.6 before I found out we have 9.4 :(

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/sql/PostgresSecondaryMetrics.scala
```diff
@@ -44,5 +46,55 @@ class PostgresSecondaryMetrics(conn: Connection) extends SqlSecondaryMetrics(con
           throw new Exception("Unexpected Postgres unique constraint violation!", uv)
       }
     }
+
+    val date = DateUtils.truncate(new Date(), Calendar.DATE)
+
+    val historyUpdated = using(conn.prepareStatement(
+      """UPDATE secondary_metrics_history
+            SET total_size = ?
+          WHERE store_id = ? AND dataset_system_id = ? AND date = ?
```
PiSoup: Since you're retrieving a new date will this ever match?

chitang: It also truncates the date

PiSoup: Actually, why do we ever want to update an entry in the history?

chitang: i thought about whether to keep the first data point in a date or the last.  The last involves more work but will give you a bit more information that may be relevant when you need it.

PiSoup: Since this is a historical record, and is going to be logging a single row per upsert, why not hold onto every datapoint in a day? I'm not sure who is the intended consumer of this record so maybe there's no value in it.

chitang: too much info isn't good.  I also heard future and history is in a loop.

chitang: i can rename it to summary if we don't want to abuse history.

znep:
It would be nice to have a view or something that can show day by day total numbers even if a dataset isn't updated that day to make querying this less convoluted and easier to use for debugging.

PR #268, opened 2019-03-23
EN-31249 Add table function secondary_metrics_history_full(date_start…

chitang:
…, date_end) for your convenience

This gives you a table with date in every day between date_start and date_end even if the underlying does not have data point.


znep:
This gives you a row for each day but, unless I'm missing something it doesn't populate that row with data if nothing has changed?  My goal is to be able to easily get a day by day record of how much data there is total by dataset, which can then be joined outside truth to get amount of data by domain over time to give reports like "top five domains adding data this month".  I think it can be done with window functions or some other way, but it isn't immediately obvious to me how to get it without some work.

chitang:
It does:
(SELECT h2.total_size
               FROM secondary_metrics_history h2
              WHERE h2.date <= t1.date
                and h2.store_id = t1.store_id
                and h2.dataset_system_id = t1.dataset_system_id
              ORDER by h2.date limit 1) as total_size

znep:
@chitang ahh, perfect, I missed that part.  Thanks!

PR #272, opened 2019-05-06
EN-32060 Pass row and row previous image to handlers.

chitang:
Include target column (in addition to source columns) in comparing row change/no change.

These two changes allow handlers to have more flexibility in determining when to recompute.

### DIFF
--- a/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
+++ b/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
@@ -123,7 +123,7 @@ class FeedbackContext[CT,CV](user: String,
     val toCompute = rows.iterator.zipWithIndex.flatMap { case (row, index) =>
       val rcis =
         perColumnDataAndDependentColumns.flatMap { case (columnData, dependentColumns) =>
-          // Don't compute if there has been to change to the dependent columns which include source columns + target column
+          // Don't compute if there has been no change to the dependent columns which include source columns + target column.
           // New value of either target or source will have the same existing if update does not contain the field.
           // There should be no change either they contain the same value or are not passed in
           if (noChange(row, dependentColumns))
--- a/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
+++ b/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
@@ -116,16 +116,14 @@ class FeedbackContext[CT,CV](user: String,
                              rows: IndexedSeq[Row[CV]],
                              targetColumns: Set[UserColumnId]): Either[ComputationFailure, Map[Int, Map[UserColumnId, CV]]] = {
     val perDatasetData = computationHandler.setupDataset(cookie)
-    val perColumnDataTargetColMap = cookie.strategyMap.toSeq.collect {
+    val perColumnDataAndTargetColId = cookie.strategyMap.toSeq.collect {
       case (targetCol, strat) if targetColumns.contains(targetCol) & computationHandler.matchesStrategyType(strat.strategyType) =>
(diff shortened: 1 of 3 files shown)


secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
```diff
@@ -116,18 +116,22 @@ class FeedbackContext[CT,CV](user: String,
...
           // don't compute if there has been to change to the source columns
-          if (noChange(row, columnData.strategy.sourceColumnIds))
+          val targetColId = perColumnDataTargetColMap(columnData)
+          // New value of either target or source will have the same existing if update does not contain the field.
+          // There should be no change either they contain the same value or are not passed in
+          if (noChange(row, columnData.strategy.sourceColumnIds :+ targetColId))
             None
           else
-            Some(computationHandler.setupCell(columnData, row.data))
+            Some(computationHandler.setupCell(columnData, row))
```
rjmac: Why push this `.data` dereference down into `setupCell`?  Also it makes the imports rather confusing, since down below in the other two classes you have both `com.socrata.datacoordinator.secondary.feedback.Row` and `com.socrata.datacoordinator.secondary.Row` in scope unqualified (the latter via wildcard-import, which grr someone's IDE was overzealous about collapsing multiple imports into a wildcard import, that is my least favorite intellij misfeature) and it confused me about how this change could even compile for a bit.

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
```diff
@@ -116,18 +116,22 @@ class FeedbackContext[CT,CV](user: String,
...
+    val perColumnData = perColumnDataTargetColMap.keys.toSeq
     val toCompute = rows.iterator.zipWithIndex.flatMap { case (row, index) =>
       val rcis =
         perColumnData.flatMap { columnData =>
           // don't compute if there has been to change to the source columns
-          if (noChange(row, columnData.strategy.sourceColumnIds))
+          val targetColId = perColumnDataTargetColMap(columnData)
+          // New value of either target or source will have the same existing if update does not contain the field.
+          // There should be no change either they contain the same value or are not passed in
+          if (noChange(row, columnData.strategy.sourceColumnIds :+ targetColId))
```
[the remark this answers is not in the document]
chitang: refactor

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
chitang: Determine when and when not to geocode is still problematic.  Currently, we have a bug where upserting the same value will erase the computed value.  Here is the path:
core upsert (unfortunately set location/point to null).  DC clears the point.  geocoding secondary doesn't even have a chance to re-compute because it detects no change for it only compares source columns excluding target column.  But if we pass in only one copy of row/data to the handler, there are times it will just accept the new target value (which is undesirable) when the first problem is fixed.

chitang: I actually have a different version that rename feedback.Row to something else.  What if we use FeedbackRow or RowWithPreviousImage?  I was just trying to minimize superficial code change.

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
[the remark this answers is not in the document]
rjmac: That doesn't change the thing I was talking about, which was the `columnData.strategy.sourceColumnIds :+ targetColId` part.  This will always produce the same result for a given column, but constructs a new version of it for each row in the column.

PR #277, opened 2019-06-03
EN-32749 Be able to suppress secondary indexing and (re)trigger index…

chitang:
…ing via noop

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Mutator.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Mutator.scala
@@ -190,6 +190,9 @@ object Mutator {
   case class DropRollup(index: Long, name: RollupName) extends Command
   case class AddComputationStrategy(index: Long, id: ColumnIdSpec, computationStrategy: ComputationStrategySpec) extends Command
   case class DropComputationStrategy(index: Long, id: ColumnIdSpec) extends Command
+  case object NoOpCommand extends Command {
+    def index: Long = 0L
+  }
 
   val AddColumnOp = "add column"
   val DropColumnOp = "drop column"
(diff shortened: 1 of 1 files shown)


coordinator/src/main/scala/com/socrata/datacoordinator/service/Mutator.scala
```diff
@@ -268,6 +272,8 @@ class Mutator[CT, CV](indexedTempFile: IndexedTempFile, common: MutatorCommon[CT
             val updateOnly = getWithStrictDefault[Boolean]("update_only", false)
             val bySystemId = getWithStrictDefault[Boolean]("by_system_id", false)
             RowData(index, truncate, mergeReplace, nonFatalRowErrorsClasses, updateOnly = updateOnly, bySystemId = bySystemId)
+          case NoOp =>
+            NoOpCommand
```
rjmac: Pass the `index` into the command here.  It doesn't look like it's needed for anything, but it's best to keep the pattern constant.

What on earth is is this for though, and how does it relate to the PR description?

chitang: There will be metadata added to pg secondary to control indexing behavior.  Initially it is going to be manually entered.  In the future, it will be driven by some other automatic mechanism. We are using the same secondary replication version update mechanism to drive index update or rebuild (like after we mess around with secondary indexes)

rjmac: Ah.  You don't actually need a no-op command for that, then.  If all you want is to poke secondaries, a mutation script with a "normal" command header and no subsequent commands will create a new data version with an empty update list that gets pushed out to secondaries.

chitang: ok, i'll try that

PR #279, opened 2019-06-04
EN-32779 Lazy indexing

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractSqlLogger.scala
```diff
@@ -172,6 +173,11 @@ abstract class AbstractSqlLogger[CT, CV](val connection: Connection,
     logLine(SecondaryReindex, messages.SecondaryReindex.defaultInstance)
   }
 
+  def secondaryAddIndex(fieldName: ColumnName) = {
+    checkTxn()
+    logLine(SecondaryAddIndex, messages.SecondaryAddIndex.defaultInstance)
```
rjmac: You're missing putting the field name on the message here

coordinatorlib/src/main/protobuf/LogData.proto
```diff
@@ -130,5 +130,9 @@ message RowsChangedPreview {
 message SecondaryReindex {
 }
 
+message SecondaryAddIndex {
+  required string fieldName = 2;
```
rjmac: 2?  I don't think it matters from a technical point of view, but I'm wondering why you picked 2 to start with.

chitang: It was copied from somewhere else and i forgot to adjust the index.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractSqlLogger.scala
```diff
@@ -175,7 +175,7 @@ abstract class AbstractSqlLogger[CT, CV](val connection: Connection,
 
   def secondaryAddIndex(fieldName: ColumnName) = {
     checkTxn()
-    logLine(SecondaryAddIndex, messages.SecondaryAddIndex.defaultInstance)
+    logLine(SecondaryAddIndex, messages.SecondaryAddIndex(fieldName.caseFolded))
```
rjmac: This one _should_ be `name`, not `caseFolded` because it'll get put back into a `ColumnName` on the other end

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDelogger.scala
```diff
@@ -183,6 +183,8 @@ class SqlDelogger[CV](connection: Connection,
           null
         case SqlLogger.SecondaryReindex =>
           Delogger.SecondaryReindex
+        case SqlLogger.SecondaryAddIndex =>
+          Delogger.SecondaryReindex
```
rjmac: This should be `secondaryAddIndex(aux)`, where that looks like:

```scala
def secondaryAddIndex(aux: Array[Byte]) = {
  val msg = messages.SecondaryAddIndex.defaultIndex.mergeFrom(aux)
  Delogger.SecondaryAddindex(ColumnName(msg.fieldName))
}
```

(+/- typos because I just entered that off the top of my head; there are lots of examples of this below)

PR #280, opened 2019-07-10
Use bulk update statements with pgsql, not just batched single updates

rjmac:
While it's not comparable to the improvement COPY gives you over INSERT,
this is somewhat faster than batched single updates.


chitang:
:shipit:

PR #281, opened 2019-07-17
Allow url type to be used as row identifier

chitang:
Probably won’t do

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
@@ -12,6 +12,8 @@ class UrlRep(val base: String) extends RepUtils with SqlPKableColumnRep[SoQLType
 
   val physColumns: Array[String] = Array(base + "_url", base + "_description")
 
+  override val keyColumns: Array[String] = Array(physColumns(0))
+
   val sqlTypes: Array[String] = Array("TEXT", "TEXT")
 
   // sub-column offsets
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
@@ -3,10 +3,10 @@ package com.socrata.datacoordinator.common.soql.sqlreps
 import java.lang.StringBuilder
 import java.sql.{PreparedStatement, ResultSet, Types}
 
-import com.socrata.datacoordinator.truth.sql.SqlColumnRep
+import com.socrata.datacoordinator.truth.sql.SqlPKableColumnRep
 import com.socrata.soql.types._
 
-class UrlRep(val base: String) extends RepUtils with SqlColumnRep[SoQLType, SoQLValue] {
+class UrlRep(val base: String) extends RepUtils with SqlPKableColumnRep[SoQLType, SoQLValue] {
 
   def representedType: SoQLType = SoQLUrl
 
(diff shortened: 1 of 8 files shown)


chitang:
It seems pretty invasive.  Is there better idea?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/json/JsonColumnRep.scala
```diff
@@ -13,6 +13,16 @@ trait JsonColumnReadRep[CT, CV] extends JsonColumnCommonRep[CT, CV] {
...
   def toJValue(value: CV): JValue
+
+  /**
+    * Output of upsert mutation are in the form:
+    *   [{"typ":"insert","id": $JVALUE,"ver":"rv-vhb7.vn49.4xy9"}]
+    * Libraries were written to only handle "id" as json string.
+    * This allows complex type like URL to have a different/simplified json format when it is used
+    * as return id value in upsert mutation.
+    */
+  def toRowIdJValue(value: CV): JValue = toJValue(value)
```
rjmac: The point of returning the row IDs here is so that user code that cares can align their idea of the value with the version that also come back (or the error, if the row update failed).  Returning only part of the value seems to defeat that purpose.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
```diff
@@ -26,6 +28,52 @@ class UrlRep(val base: String) extends RepUtils with SqlColumnRep[SoQLType, SoQL
...
+    val url = literal.asInstanceOf[SoQLUrl]
+    url match {
+      case SoQLUrl(Some(url), Some(description)) =>
+        s"(${base}_url = ${sqlescape(url)} AND ${base}_description = ${sqlescape(description)})"
+      case SoQLUrl(Some(url), None) =>
+        s"(${base}_url = ${sqlescape(url)})"
+      case SoQLUrl(None, Some(description)) =>
+        s"(${base}_description = ${sqlescape(description)})"
+      case SoQLUrl(None, None) =>
+        s"(${base}_url is null and ${base}_description is null)"
```
rjmac: It seems weird to have this one case care about subcolumns being null, and not the other two.

rjmac: It's also weird that this is incongruent with `sql_in`'s behavior.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/json/JsonColumnRep.scala
rjmac: Note BTW that no user code cares about this value, because it's eaten by core.  If we change it so that it's not guaranteed to be a string, we don't have to worry about breaking anyone other than ourselves.

chitang:
other options are
1. use both url.url and url.description to match records
2. deprecate url as row identifier (this is product decision though)

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/json/JsonColumnRep.scala
chitang: changed.  Will require change in soda-java client/core reference.  Wouldn't it affect people who use soda-java client?  This use case is small enough that I guess breaking/requiring update is fine.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/UrlRep.scala
chitang: fixed

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/json/JsonColumnRep.scala
rjmac: As far as I'm aware, there is no way for people who use the soda-java client to get this response.

PR #296, opened 2019-12-17
EN-37185: Fix deleting rows with a `null` value

rjmac:
This would 500.  Now it just returns "no such row to delete", which I
think is reasonable.


chitang:
:shipit:

PR #305, opened 2020-01-24
EN-37610 collocation capacity overflow

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/config/SecondaryConfig.scala
```diff
@@ -5,12 +5,12 @@ import com.typesafe.config.Config
...
 
 trait StoreConfig {
-  val storeCapacityMB: Int
+  val storeCapacityMB: Long
   val acceptingNewDatasets: Boolean
 }
 
 class StoreConfigCfg(config: Config, root: String) extends ConfigClass(config, root) with StoreConfig {
-  val storeCapacityMB = getInt("storeCapacityMB")
+  val storeCapacityMB = getInt("storeCapacityMB").toLong
```
PiSoup: Could this just be `getLong` or does that method not exist?

chitang: That method doesn't exist yet.  I open another PR to provide that in the library.  But for hotfix-ish purpose.  I rather do this as a baby step.

PR #309, opened 2020-08-07
Consolidate multiple row-update-only data versions into one big data …

rjmac:
…version

In particular you can get a ton of such versions when the feedback
secondaries discover a big change.  This has the potential to make things
MUCH faster on large updates to large datasets even, because it can mean
the feedback secondary has half a hope of triggering the "make an indexless
copy" logic.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
@@ -219,6 +219,7 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
     // A series of events is consolidatable if it is an update which only
     // changes row data, which is to say if it has the form:
     //    RowsChangedPreview
+    //    zero or one Truncated
     //    zero or more RowDataUpdated
     //    LastModifiedChanged
     def consolidatable(it: BufferedIterator[Delogger.LogEventCompanion]): Boolean = {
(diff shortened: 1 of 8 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -185,6 +183,176 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
+        val data = (effectiveStartingDataVersion to endingDataVersion).iterator.flatMap { i =>
+          new Iterator[Delogger.LogEvent[CV]] {
+            var done: Boolean = false
+
+            val eventsRaw = rs.open(delogger.delog(i))
+            val events = eventsRaw.buffered
+            if(!events.hasNext || events.head.companion != Delogger.RowsChangedPreview) {
+              throw ResyncSecondaryException("Consolidation saw the log change?!  First item was gone or not RowsChangedPreview!")
+            }
+            if(events.next().asInstanceOf[Delogger.RowsChangedPreview].truncated) {
```
rozap, line 282: i don't think i understand this cast..

rjmac: `truncated` is a field on `Delogger.RowsChangedPreview`, which is a subclass of `Delogger.Event` - in the previous conditional we've checked that it both exists and _is_ a RCP, so the downcast is safe.

rjmac: (this would be clearer to write with `nextOption` and a `match`, but `nextOption` doesn't actually exist on `Iterator`)

PR #313, opened 2020-09-08
EN-40823: add secondary instruction - delete index

chitang:
Add index_directives table in migration


langlk:
This looks good to me, but you might want to have someone a little more familiar with data-coordinator take a look

PR #315, opened 2020-09-17
Chi/en 40823/delete index3

chitang:


vickingan (approved):
Looks fine to me but would probably also be good to have someone with more experience in this area look at it as well.

PR #320, opened 2020-10-05
Allow empty upserts to be consolidated.

rjmac:
Turns out this is important when you have large datasets being delta-
updated regularly!


vickingan (approved):
👍 LGTM

PR #324, opened 2020-11-18
Don't _just_ internal-error when collocation fails unexpectedly.

rjmac:
Say something too!


znep (approved):
I assume there is already enough context about the job from the rest of the logging?

rjmac:
Nope!  But there are several things that can go wrong and I haven't the slightest idea which one of them it is.

PR #329, opened 2020-11-20
Handle ancient timestamps better

rjmac:
This makes me sad, because we're now sending timestamps around as
strings, but java.sql.Timestamp doesn't deal with pre-Gregorian
dates very well.  But at least it ensures we'll read out the same
values we wrote in...


chitang:
:shipit:  😢

PR #330, opened 2020-11-20
Get timestamps as strings, take 2

rjmac:
soql-postgres-adapter didn't like the previous one, where the select
list turned the timestamp into an ISO string.  Instead, we'll understand
Postgres's format.


chitang:
:shipit:

PR #331, opened 2020-12-04
Fix timestamp parsing on extreme (>4-digit-year) timestamps

rjmac:
Also remove the Java timestamp helper (because its purpose for
existing -- which was never real in the first place -- is now
thoroughly gone.


catstavi:
is this something we could have a fun little unit test for?

catstavi (approved):

PR #333, opened 2021-04-13
EN-45573 SoQLJson support

rozap:
* initial support for soql json data type
* depends on https://github.com/socrata-platform/soql-reference/pull/183
* to be merged after https://github.com/socrata-platform/soql-postgres-adapter/pull/393


saraid (approved):
No idea about whether or not this plumbs everything together correctly, and I wish there were some tests kicking around, but as well as I can follow it, it seems fine.

PR #335, opened 2021-06-28
Option to co-secondary-stores on dataset create

chitang:


coordinator/src/main/scala/com/socrata/datacoordinator/resources/collocation/SecondaryManifestsCollocateResource.scala
```diff
@@ -42,12 +42,12 @@ case class SecondaryManifestsCollocateResource(storeGroup: String,
...
   }
 
- def doCollocationJob(jobId: UUID,
-                      storeGroups: Set[String],
-                      request: CollocationRequest,
-                      explain: Boolean): Either[ErrorResult, CollocationResult] = {
+  def doCollocationJob(jobId: UUID,
+                       storeGroups: Set[String],
+                       request: CollocationRequest,
+                       explain: Boolean): Either[ErrorResult, CollocationResult] = {
```
MarkWillisford, line 48: This might be an obvious answer, but what are the meanings of these 1 space indentations? I couldn't find any explanation in the Scala docs but I'm not very familiar with the applicable syntax rules

chitang: The def has only 1 space.  It should be 2.  The parameters are aligned to each other not a multiple of ident x space.

MarkWillisford (approved):

PR #338, opened 2021-12-07
EN-49288: Bump soql-reference

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/SoQLRowLogCodec.scala
```diff
@@ -118,6 +118,8 @@ object SoQLRowLogCodec extends SimpleRowLogCodec[SoQLValue] {
         target.writeStringNoTag(id)
       case SoQLNull =>
         target.writeRawByte(-1)
+      case SoQLInterval(_) =>
+        ???
```
MarkWillisford, line 122: So this is an undefined placeholder . . .   I'm assuming we can never actually get here?

chitang: Interval is 1/3 of a data type.  We use it to handle query function argument but it cannot get out to result set.

chitang: And you can't use that as column data type.

MarkWillisford (approved):

PR #344, opened 2022-02-15
EN-50425: Consolidate rollups

rjmac:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -216,34 +220,67 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
     // A series of events is consolidatable if it is an update which only
-    // changes row data, which is to say if it has the form:
+    // changes row data or rollups, which is to say if it has the form:
     //    RowsChangedPreview
     //    zero or one Truncated
     //    zero or more RowDataUpdated
     //    LastModifiedChanged
     // or
+    //    zero or more intermingled RollupCreatedOrUpdated and RollupDropped
     //    LastModifiedChanged
```
catstavi, line 238: ok so I walked through this (old) logic, and while the set above makes sense, the thing I don't get here is that rather than "a single LastModifiedChanged" the code seems to actually be saying "a single LastModifiedChanged followed by exactly one other LogEvent of any type" ?

probably missing something but just jotting down my confusion before I switch over to the new logic

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -216,34 +220,67 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
-        it.next()
-      }
-      if(!it.hasNext || it.next() != Delogger.LastModifiedChanged) {
-        return false
+      if(!it.hasNext) {
+        return Set.empty
       }
 
-      !it.hasNext
+      it.next() match {
```
catstavi, line 256: 👍 read through this logic and looks to me like match for the updated comment above. Only thing I'm still confused about is the "single LastModifiedChanged" log event scenario (it looks like there must be a hasNext of any type for that to work?)

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -256,109 +293,228 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
...
+
+        val (effectiveStartingDataVersion, state) =
+          (startingDataVersion to endingDataVersion).foldLeft((startingDataVersion, Unknown:State)) { (acc, i) =>
+            val (currentStartingDataVersion, state) = acc
+
+            def rcpOfState() =
+              state match {
+                case Unknown => Delogger.RowsChangedPreview(0, 0, 0, false)
+                case ConsolidatingRows(rcp) => rcp
+                case ConsolidatingRollups => throw new ResyncSecondaryException("Was consolidating rows, now rollups?")
```
catstavi, line 309: are these two not allowed as sequential states?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
rjmac: So, LastModifiedChanged should always be the last event in a version.  If it's not something's up and we bail (== don't consolidate out of caution).  What the new code at lines 242-254 is doing is saying "If there is a first event and that first event is LMC and there are no more events, we can consolidate with anything (hm 248 should be "return Consolidatable.all", I'll fix that) otherwise we can consolidate with nothing.  If there was no first event something unexpected is happening and we can consolidate with nothing".

Which means that if we get here to 256, we know that there is a first event and it is _not_ LMC so we can split down the two different types of consolidation that we're willing to perform.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
rjmac: So no, it says "a single LMC followed by exactly nothing" - `it.head` on line 228 doesn't advance the iterator, so the subsequent `it.next()` just pops the LMC out in order to let us see if it's the end of the version.

catstavi: ok, I think that was my confusion-- I was thinking that for an iteratior that contains for example [LMC], if head == LMC, then hasNext would be false (because of nothing being after LMC)

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
catstavi: This was just the same misunderstanding on my part as above

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
catstavi, line 309: ok, this is most likely just another point of misunderstanding for me, but is the text here in reverse? (ie: should it actually be "Was consolidating rollups, now rows?"

Because it seems like we'll hit this case if the "state" from the previous accumulation is ConsolidatingRollups (meaning we *were* dealing with rollups stuff), but we called this function because we got a row-related event from the delogger?

catstavi:
ok! I have read this through and I think I get the picture of what's going on! We're introducing a new sequence of log events that can be consolidated (a sequence of rollup-y events + ldm). When we detect that, we consolidate the iterator through the versions with a consolidated iterator.

I don't have a good picture for how this piece fit into everything else going on (what's being done with the iterator?), but that's alright. There was nothing I wrong I was able to catch, though I will wait for an anwer to that question about the ResyncSecondaryException above.

If you want a more knowledgeable look at this, we can tag chi!

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
rjmac: Hm yeah, I think you're right

rjmac:
Yep, you've got it exactly!  Will clarify those resync exception messages.

catstavi (approved):

PR #347, opened 2022-03-03
EN-51020: Missing invalid rollups

chitang:


chitang:
Previously, rollups using :id are missing truth if they were created with working copies and then published.  Secondaries still maintain these "missing" rollups which are now orphaned.
Rollup goes through different processing when different event happen.  On publish, rollups go through some kind of validation but the setup was wrong.  It erroneously prepend system columns with _, causing :id to become _:id.  This PR does a few things:

1. stop turning :id into _:id
2. make create/update rollup go through the same validation as in publish
3. when validation does not pass validation, write drop rollup to the truth log.  Previously, it only dropped the rollup from truth metadata.

catstavi (approved):
thank you so much for the clear description of the changes! It helped so much for reading the code to validate that everything matched up with your goals, and I also feel like I already have a much better understanding of rollup process and what's happening here!

PR #348, opened 2022-03-28
EN-51706: Handle queries containing multiple tables in rollup validation

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/DatabaseAccessors.scala
```diff
@@ -435,18 +439,40 @@ object DatasetMutator {
...
+              getPrefixedDsContext(copyInfo)
+            }
+            ctx match {
+              case Some(context) =>
+                acc + (tn -> context)
+              case None =>
+                acc
+            }
+          }
+          analyzer.analyzeBinary(parsed)(contexts)
```
catstavi, line 475: Ok, so, checking my understanding:
* we need to pass all the various tables (not just the primary table) in context, so that rollups can work cross-table
* in order to extract all the table names, we need to work with the query in binaryTree format instead of soql format

* but in the end, we're still calling analyze with the context to validate
yes?

chitang: That's right

catstavi (approved):

PR #350, opened 2022-04-13
Resync API

zachlefevre:


rozap (approved):
:ship: we can add the not-found case later

rozap (changes requested):
wait actually - we should maybe add that case first. i'm not sure what the behavior will be if we add random stuff to the secondary manifest this way...

one weird case: there's dataset `A` and it gets soft deleted, then resync gets called on `A`...that may do weird stuff. so we may want to do the lookup of the thing and make sure it's in a state that can be resynced, before we poke the manifest.

zachlefevre:
We'll never add a record to the DB that isn't already there, since this is an update.
We can only perform the update if pending_drop is false. I assume (and have not tested) the secondary will stop receiving events and resyncs if the dataset is pending a drop, so updating the manifest will have no effect.

rozap:
yea, it wouldn't be a problem in hard deletes, but should we have a `WHERE not pending_drop` in that update query?

rozap (approved):

PR #351, opened 2022-04-15
Add column rollup_map.raw_soql

chitang:


coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20220415-add-column-rollup-raw-soql.xml
```diff
@@ -0,0 +1,16 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+    <changeSet author="Chi" id="20220415-add-column-rollup-raw-soql">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="rollup_map" columnName="raw_soql"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="rollup_map">
+            <column name="raw_soql" type="varchar"/>
```
saraid, line 13: Vaguely surprised you're using `varchar` here. Or does that no longer have a reachable max length?

chitang:
Pretty insignificant and for convenience at very little cost.
user_soql might be a better name

saraid (approved):
I don't know enough about this to have a problem with it.

chitang:
The field rollup_map.soql is a transformed soql with internal column ids which is not for human readability.  In theory, we have code to convert this soql back and forth.  But it is a difficult process.  Bugs can also get in the way.  It would be nice to save a copy of the original soql given by the user.

coordinatorlib/src/main/resources/com.socrata.datacoordinator.truth.schema/20220415-add-column-rollup-raw-soql.xml
chitang: This field mirrors rollup_map.soql which is varchar.

PR #353, opened 2022-05-06
Fix crash when on null SoQLJson cells

rozap:
* if you upserted a null value in a soql json
  column, it would crash


rozap:
@chitang gave :shipit: over slack

PR #356, opened 2022-05-09
EN-52985: Custom index support

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/id/sql/package.scala
```diff
@@ -73,4 +73,21 @@ package object sql {
...
+  }
+
+  implicit class IndexIdGetter(val __underlying: ResultSet) extends AnyVal {
+    private def IndexIdify(x: Long) =
+      if(__underlying.wasNull) IndexId.Invalid
+      else new IndexId(x)
+    def getIndexId(col: String): IndexId =
+      IndexIdify(__underlying.getLong(col))
+    def getIndexId(idx: Int): IndexId =
+      IndexIdify(__underlying.getLong(idx))
```
rjmac, line 91: Don't much like this "Invalid" thing, but it doesn't look like it can happen in practice given the queries, and it does, alas, follow the convention elsewhere.

rjmac (approved):

PR #358, opened 2022-05-09
Import the v3 instead of the v2 versions of various rojoma-json things

rjmac:
I'm pretty sure these were all bugs that just happened to go unnoticed
since they're all exceptions.


chitang (approved):
maybe bump version?

PR #363, opened 2022-05-31
EN-53570: Be able to exclude columns from the search index

chitang:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/AbstractSqlLogger.scala
```diff
@@ -181,7 +181,11 @@ abstract class AbstractSqlLogger[CT, CV](val connection: Connection,
       case Some(JBoolean(b)) => b
       case _ => false
     }
-    logLine(IndexDirectiveCreatedOrUpdated, LogData.IndexDirectiveCreatedOrUpdated(convert(info.unanchored), enabled))
+    val search = directive.get("search") match {
+      case Some(JBoolean(b)) => Some(b)
+      case _ => Some(true)
```
catstavi, line 186: should this be None?

chitang: This is kinda intentional just in case the search key isn't a boolean.

catstavi (approved):

PR #364, opened 2022-06-09
Fix ensuring dataset-in-same-secondaries when there's multiple colloc groups

rjmac:
Before, we were finding the secondaries of dataset and then finding
any that were not (in our target collocation group and accepting new
datasets) but if the _DEFAULT_ meta-group has multiple groups in it,
then this will always fail when it asks "is this secondary from group B
in the set of datasets in group A that are still accepting datasets?"

So, when doing this check, limit ourselves to the secondaries that are
actually in the group under consideration.

### DIFF
--- a/coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
+++ b/coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
@@ -506,7 +506,7 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
     val preferredSecondaries = secondariesLike match {
       case Some(datasetInternalName) =>
         val candidateSecondariesInGroup = secondaryGroup.instances.filter(_._2.acceptingNewDatasets).keySet
-        val secs = collocationProvider.collocator.secondariesOfDataset(datasetInternalName)
+        val secs = collocationProvider.collocator.secondariesOfDataset(datasetInternalName).filter(secondaryGroup.instances.keySet)
         secs.find(!candidateSecondariesInGroup.contains(_)) match {
           case Some(sec) =>
             throw new Exception(s"Secondary ${sec} like ${datasetInternalName.underlying} is no longer accepting datasets" )


coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -506,7 +506,7 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
     val preferredSecondaries = secondariesLike match {
       case Some(datasetInternalName) =>
         val candidateSecondariesInGroup = secondaryGroup.instances.filter(_._2.acceptingNewDatasets).keySet
-        val secs = collocationProvider.collocator.secondariesOfDataset(datasetInternalName)
+        val secs = collocationProvider.collocator.secondariesOfDataset(datasetInternalName).filter(secondaryGroup.instances.keySet)
```
housepage, line 509: Should we use `candidateSecondariesInGroup` since that does not contain secondaries that aren't accepting new datasets?

housepage (changes requested):
This change make sense based on what you described. Had one question.

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
housepage: Oh nm. I see that we're comparing the two below and generating an error message.

rjmac: No, this needs to include secondaries that aren't accepting new datasets.

housepage (approved):
:sheep: :it:

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
rjmac: (because if we exclude those we'll never find a secondary that is genuinely not accepting new datasets on the next line)

catstavi (approved):

PR #371, opened 2022-07-26
Use SKIP LOCKED for postgresql dequeue instead of a subselect

rjmac:
This appears to have exactly the behavior we want with respect to
number of rows locked by the query.


catstavi (approved):
neat!

PR #372, opened 2022-08-15
EN-55251: Prevent dataset leak in pg secondary

chitang:


rjmac (approved):
:+1: though I think I'd like an even bigger margin?  If we don't have any secondary-watcher workers available for an entire day we have big problems, but it's not _completely_ inconceivable.

PR #374, opened 2022-09-23
Deprecate Jenkins force build parameter

ds2643:
Previously, cut jobs on Jenkins used in the release process were executed only when projects accumulated commits since the prior release. However, the introduction of the Aqua constraint requires releasing all projects each release cycle. Accordingly, the concept of force building-- executing a cut when no changes have happened-- should be default behavior. That is, projects should always build and deploy to RC when the cut job runs.

To address this problem, this commit removes the FORCE_BUILD parameter from this project's Jenkins cut job. As a consequence, the project will build regardless of how many commits follow the previous release, aiding ergonomics.

Relates to [EN-55643](https://socrata.atlassian.net/browse/EN-55643).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,6 @@ pipeline {
   }
   parameters {
     booleanParam(name: 'RELEASE_CUT', defaultValue: false, description: 'Are we cutting a new release candidate?')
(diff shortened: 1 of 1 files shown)


sarah-salinas (approved):
The changes look good.

PR #377, opened 2022-10-19
Cross-truth dataset mover

rjmac:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
```diff
@@ -806,10 +806,10 @@ trait BasePostgresDatasetMapWriter[CT] extends BasePostgresDatasetMapReader[CT]
...
       stmt.setLong(1, newDatasetInfo.nextCounterValue)
       stmt.setLong(2, newDatasetInfo.latestDataVersion)
-      stmt.setDatasetId(3, newDatasetInfo.systemId)
-      stmt.setString(4, newDatasetInfo.localeName)
-      stmt.setBytes(5, newDatasetInfo.obfuscationKey)
-      stmt.setString(6, newDatasetInfo.resourceName.orNull) // resource_name is a nullable column
+      stmt.setString(3, newDatasetInfo.localeName)
+      stmt.setBytes(4, newDatasetInfo.obfuscationKey)
+      stmt.setString(5, newDatasetInfo.resourceName.orNull) // resource_name is a nullable column
+      stmt.setDatasetId(6, newDatasetInfo.systemId)
```
saraid, line 812: I don't understand the purpose of this change.

dataset-mover/src/main/scala/com/socrata/datacoordinator/mover/Main.scala
```diff
@@ -0,0 +1,490 @@
...
+            if(chunk.length > buffer.length) {
+              toCopyIn.writeToCopy(chunk, 0, chunk.length)
+              written += chunk.length
+            } else {
+              System.arraycopy(chunk, 0, buffer, bufferEnd, chunk.length)
+              bufferEnd += chunk.length
+            }
+            val newTotal = total + written
+            val now = System.nanoTime()
+            if(now - lastWrote > 1000000000L) {
```
saraid, line 108: I think this is the weirdest way to express "1 second has passed" I've ever seen, but I guess it works.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/metadata/sql/PostgresDatasetMap.scala
rjmac: This function was previously unused and was broken.

saraid: 🤦

saraid (approved):

dataset-mover/src/main/scala/com/socrata/datacoordinator/mover/Main.scala
rjmac: Fun fact: System.nanoTime() has no epoch, so only differences between two values acquired within the same process has a meaning (in particular: it's not guaranteed that a later timestamp will be "greater than" an earlier timestamp, only that the difference between them will be the difference in nanoseconds up to wraparound)

PR #378, opened 2022-10-21
Reset secondary replay count when updating retries

rjmac:


saraid (approved):
What could possibly go wrong?

PR #383, opened 2022-12-01
Do normal secondary playbacks without holding a truth transaction

rjmac:
One of our persistent problems on the truthstores is related to long-running transactions preventing postgresql from vacuuming its transaction log.  In particular, there is a _lot_ of churn on secondary-manifest, both due to updates incrementing latest_data_version when they complete and the keepalive poking `claimed_at` every couple of minutes while a playback is in progress.

This prevents normal (non-resync) secondary playback from holding a transaction the entire time.  I'm hoping this will improve things because updates only touch secondary_manifest right at the very end of their transaction, whereas secondary playback touches it at the beginning _and_ the end, thus (my hypothesis goes) preventing secondary_manifest _in particular_ from being cleaned in a timely fashion.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDelogger.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDelogger.scala
@@ -10,6 +10,7 @@ import com.socrata.datacoordinator.truth.RowLogCodec
 import com.socrata.datacoordinator.truth.loader._
 import com.socrata.datacoordinator.util.{CloseableIterator, LeakDetect}
 import com.socrata.soql.environment.ColumnName
+import org.slf4j.{Logger, LoggerFactory}
 import messages.LogData
 
 class SqlDelogger[CV](connection: Connection,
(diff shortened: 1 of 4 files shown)


saraid (approved):

coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDelogger.scala
```diff
@@ -19,24 +20,20 @@ class SqlDelogger[CV](connection: Connection,
...
-
   private val rs = new ResourceScope
   var _queryStmt: PreparedStatement = null
   var _queryTypesStmt: PreparedStatement = null
 
   def queryStmt = {
     if(_queryStmt == null) {
-      _queryStmt = rs.open(connection.prepareStatement("select subversion, what, aux from " + logTableName + " where version = ? order by subversion"))
-      _queryStmt.setFetchSize(1)
+      _queryStmt = rs.open(connection.prepareStatement("select subversion, what, aux from " + logTableName + " where version = ? and subversion = ? + 1"))
```
rlvoyer, line 29: Why did you remove the `order by subversion` clause here?

PR #385, opened 2022-12-23
Include truncated data in the logged "rows deleted" line

rjmac:
Also include the time to do the truncation.


rlvoyer (approved):
🏅 Thanks for the quick fix!

PR #386, opened 2023-01-04
Remove retry counts from secondarylib-feedback

rjmac:
Replay requests are counted DCside and will mark the dataset broken over there automatically, so there's no need to count them on this side.

### DIFF
--- a/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
+++ b/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
@@ -63,8 +63,6 @@ class FeedbackContext[CT,CV](user: String,
   private def success(current: CookieSchema): Success = {
     Success(currentCookie.copyCurrent(
       current = current,
-      computationRetriesLeft = computationRetryLimit,
-      dataCoordinatorRetriesLeft = dataCoordinatorRetryLimit,
       resync = false,
       errorMessage = None
     ))
@@ -77,8 +75,6 @@ class FeedbackContext[CT,CV](user: String,
   private def replayComputation(reason: String, resync: Boolean): ReplayLater = {
     ReplayLater(reason, { feedbackCookie =>
       feedbackCookie.copyCurrent(
-        computationRetriesLeft = feedbackCookie.current.computationRetriesLeft - 1,
-        dataCoordinatorRetriesLeft = dataCoordinatorRetryLimit,
         resync = resync,
         errorMessage = Some(reason))
     })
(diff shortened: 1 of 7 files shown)


secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackContext.scala
```diff
@@ -77,8 +75,6 @@ class FeedbackContext[CT,CV](user: String,
   private def replayComputation(reason: String, resync: Boolean): ReplayLater = {
     ReplayLater(reason, { feedbackCookie =>
```
rjmac, line 76: This is turned into an exception over here

https://github.com/socrata-platform/data-coordinator/blob/main/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackSecondary.scala#L167

and that exception is caught and turned into a limited-retry here:

https://github.com/socrata-platform/data-coordinator/blob/main/secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala#L112

saraid: Fun fact, if you use a sha instead of `main`, you get this:

https://github.com/socrata-platform/data-coordinator/blob/cff0665bf309bf55f6c21c7c7124baf2796dd8fd/secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackSecondary.scala#L167

https://github.com/socrata-platform/data-coordinator/blob/cff0665bf309bf55f6c21c7c7124baf2796dd8fd/secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala#L112

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackCookie.scala
```diff
@@ -55,8 +51,6 @@ case class CookieSchema(dataVersion: DataVersion,
                         columnIdMap: Map[UserColumnId, ColumnId],
                         strategyMap: Map[UserColumnId, ComputationStrategyInfo],
                         obfuscationKey: Array[Byte],
-                        computationRetriesLeft: Int,
```
catstavi, line 58: is this case class also what's used for parsing/writing to the cookie field in the secondary_manifest? (meaning: does removing this mean we'll stop writing it, and will we still read cookies that have those fields ok, just ignore them?)

saraid (approved):

secondarylib-feedback/src/main/scala/com/socrata/datacoordinator/secondary/feedback/FeedbackCookie.scala
rjmac: Yes and yes, respectively - it's standard AutomaticJsonCodecBuilder thing.

PR #394, opened 2023-02-06
All secondaries but pg should not respect collocation

zachlefevre:


rlvoyer (approved):
There will be a corresponding apps-marathon change, too?

PR #395, opened 2023-02-06
Collocation checking occurs over groups, not individual instances

zachlefevre:


configs/sample.conf
```diff
@@ -33,7 +33,7 @@ com.socrata.coordinator.common = {
           }
         }
       }
-      archival {
+      archival-group {
```
zachlefevre, line 36: These are local-only and useful for testing

coordinator/src/main/scala/com/socrata/datacoordinator/service/Main.scala
```diff
@@ -85,7 +85,7 @@ class Main(common: SoQLCommon, serviceConfig: ServiceConfig) {
                 u.secondaryManifest.addDataset(storeId, datasetId)
                 // great, now do collocation stuff...
                 val collocatedDatasets =
-                  if (coordinator.secondaryGroupConfigs(storeId).respectsCollocation)
+                  if (coordinator.secondaryGroupConfigs(storeGroup).respectsCollocation)
```
zachlefevre, line 88: This is the crux of it. We needed to pass the store group through the stack to here.

rlvoyer (approved):
LGTM

Is there a Jira ticket associated w/ the change? If so, can you add it to the PR?

PR #396, opened 2023-03-16
Make the mutator manage a separate zookeeper registration

rjmac:
* When it reaches some configurable limit, disable the registration to force load to some other instance
* When it then drops below some configurable limit, reenable the registration
* Both numbers are 15 by default, which is where it started rejecting requests before.
* Also make it actually configurable via env var at runtime


saraid (approved):

saraid (approved):
:logs:

PR #399, opened 2023-03-21
EN-58705: soql-reference version bump for column mapping fix

PaulMTT:
Just need to propagate the soql-reference changes here, as we did with query-coordinator


rlvoyer (approved):
LGTM

PR #402, opened 2023-06-22
EN-60867: scan for consolidatability from the current position

rjmac:
Not the start of the job

Before:
  on every iteration of the loop we look to see if the _job_ starts
  with consolidated versions, then (after the first iteration) we
  discard them and just play back a single version.

Now:
  on every iteration of the loop we look to see if the _remaining range_
  of versions within the job is consolidatable, and if it is, consolidate
  them


rlvoyer (approved):
Subtle bug. Good catch.

PR #404, opened 2023-07-24
When initiating a data version playback, prefetch created rollups

rjmac:
And send them to the secondary as part of the initial "version" call.


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/SqlDelogger.scala
```diff
@@ -46,6 +46,22 @@ class SqlDelogger[CV](connection: Connection,
       using(stmt.executeQuery())(versionFromResultSet)
     }
 
+  def findCreateRollupEvents(fromVersion: Long, toVersion: Long): Seq[Delogger.RollupCreatedOrUpdated] =
+    using(connection.prepareStatement("select aux from " + logTableName + " where version >= ? and version <= ? what = ? order by version, subversion")) { stmt =>
```
saraid, line 50:
```suggestion
    using(connection.prepareStatement("select aux from " + logTableName + " where version >= ? and version <= ? and what = ? order by version, subversion")) { stmt =>
```
Surely?

rjmac: Surely!

saraid (approved):
I see no problems.

PR #413, opened 2023-09-27
EN-62834: Bring back exlusive-lock-on-resync

rjmac:
At least for non-feedback secondaries


saraid (approved):
probably fine

rlvoyer (approved):
Looks sensible to me, too.

PR #418, opened 2023-10-12
Allow creating secondary watcher app without needing a typesafe config

zachlefevre:
The Redshift secondary does not use typesafe config. It uses the quarkus config system.
This change allows us to construct a secondary watcher app without relying on a typesafe config.

### DIFF
--- a/secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
+++ b/secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
@@ -502,7 +502,6 @@ object SecondaryWatcherApp {
   type NumWorkers = Int
   def apply
     (dsInfo: Managed[DSInfo], reporter: Managed[MetricsReporter], curator: Managed[CuratorFramework])
-    (secondaryConfig: SecondaryConfig)
     (secondaryWatcherConfig: SecondaryWatcherAppConfig)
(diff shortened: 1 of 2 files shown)


PaulMTT (approved):

secondarylib/src/main/scala/com/socrata/datacoordinator/secondary/SecondaryWatcher.scala
```diff
@@ -497,15 +499,14 @@ object SecondaryWatcher {
...
-  def apply(secondaryProvider: Config => Secondary[SoQLType, SoQLValue]) {
-    val rootConfig = ConfigFactory.load()
-    val config = new SecondaryWatcherConfig(rootConfig, "com.socrata.coordinator.secondary-watcher")
-    PropertyConfigurator.configure(Propertizer("log4j", config.log4j))
-
+  type NumWorkers = Int
+  def apply
+    (dsInfo: Managed[DSInfo], reporter: Managed[MetricsReporter], curator: Managed[CuratorFramework])
+    (secondaryWatcherConfig: SecondaryWatcherAppConfig)
+    (secondaries: Map[String, (Secondary[SoQLType, SoQLValue], NumWorkers)]
```
rjmac, line 506: Why are the first three parameters all Manageds, and why three parameter lists?

rjmac (approved):

PR #431, opened 2023-12-06
Don't assume tables have the same column-names in the dataset content…

rjmac:
…s copier

Because it's not true if we're switching column naming schemes.


saraid (approved):
Huh. Interesting.

PR #442, opened 2024-04-30
New-analyzer rollups and the mover

rjmac:
Not at all sure I like this yet.

At the very least it still needs:
 * an index to make the "find associated rollups" query efficient

But also I'm starting to wonder if NA rollups should refer to internal names at all...


rjmac:
Changed the representation of rollups so that this wasn't necessary anymore.

PR #443, opened 2024-05-01
EN-66367: Make datasets' resource name required and unique

rjmac:


coordinatorlib/src/main/scala/com/socrata/datacoordinator/truth/loader/sql/messages/FromProtobuf.scala
```diff
@@ -55,7 +55,7 @@ object FromProtobuf {
        nextCounterValue = di.nextCounterValue,
        localeName = di.localeName,
        obfuscationKey = di.obfuscationKey.toByteArray,
-       resourceName = di.resourceName
+       resourceName = DatasetResourceName(di.resourceName.getOrElse(sys.error("Dataset info without resource name")))
```
saraid, line 58: Does `sys.error` explode? Or at least not return a String?

rjmac: Yes, it explodes.

saraid (approved):
I'm a little baffled by the ProtoBuf interface, but I assume the compiler would have caught an egregious issue, so I'm approving.

rjmac:
The thing with the protobuf interface is that I'm not sure you can change a field from "optional" to "required" backwards-compatibly, and I don't want to invalidate existing log entries, so I'm just leaving it optional in the protobuf spec.  Not super happy, but at least the weirdness is confined to ToProtobuf/FromProtobuf.

PR #450, opened 2024-05-29
Log the difference between next_retry and "now" in the "got a job" me…

rjmac:
…ssage


saraid (approved):
I don't remember if next_retry has a non-null constraint or not. Might be worth null-checking.

rjmac:
> I don't remember if next_retry has a non-null constraint or not. Might be worth null-checking.

It does

saraid (approved):

PR #452, opened 2024-06-12
print out the whole error if e.getMessage is returning null

dnabilali:


dnabilali:
I'm getting an error related to the redshift secondary where e.getMessage is returning null.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -655,7 +655,8 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
         return actions
       } catch {
         case e: Throwable =>
-          logger.info("Rolling back to end transaction due to thrown exception: {}", e.getMessage)
+          val msg = Option(e.getMessage).getOrElse(e.toString)
+          logger.info("Rolling back to end transaction due to thrown exception: {}", msg)
```
rjmac, line 659:
```scala
        case e: Throwable =>
          logger.info("Rolling back to end transaction due to thrown exception:", e)
```
...will get you a stack trace in addition to the message, if you want that.

rjmac (approved):
:+1: if you don't want the stack trace, an alternative if you do.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
dnabilali: just to make sure we're on the same page: you think it's okay to log e for all services going through data-coordinator?

coordinatorlib/src/main/scala/com/socrata/datacoordinator/secondary/PlaybackToSecondary.scala
```diff
@@ -655,7 +655,7 @@ class PlaybackToSecondary[CT, CV](u: PlaybackToSecondary.SuperUniverse[CT, CV],
         return actions
       } catch {
         case e: Throwable =>
-          logger.info("Rolling back to end transaction due to thrown exception: {}", e.getMessage)
+          logger.info("Rolling back to end transaction due to thrown exception: {}", e)
```
rjmac, line 658: No placeholder if you want the stack trace (I'm not actually 100% sure what this will do, but I _think_ it'll log `e.toString`).  slf4j treats an extra throwable argument specially.

dnabilali: oh okay! thank you!

rjmac (approved):

PR #458, opened 2024-08-13
EN-68076: Consistently remove NUL characters from text values

rjmac:
Before, we were doing this on _inserts_ of specifically _text_ values. Now we do it everywhere a user-provided datum is a text value.

### DIFF
--- a/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/TextRep.scala
+++ b/coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/TextRep.scala
@@ -52,7 +52,7 @@ class TextRep(val base: String) extends RepUtils with SqlPKableColumnRep[SoQLTyp
 
   def csvifyForInsert(v: SoQLValue) = {
     if(SoQLNull == v) Seq(None)
-    else Seq(Some(v.asInstanceOf[SoQLText].value.replace("\u0000", "")))
+    else Seq(Some(v.asInstanceOf[SoQLText].value))
   }
 
   def prepareInsert(stmt: PreparedStatement, v: SoQLValue, start: Int): Int = {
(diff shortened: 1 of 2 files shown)


coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/RepUtils.scala
```diff
@@ -38,4 +38,8 @@ abstract class RepUtils {
     }
     sb.append(q)
   }
+
+  def cleanText(s: String): String =
+    if(s.indexOf('\u0000') == -1) s // fastpath: no NULs to remove
```
zachlefevre, line 43: Why 4 0s? Isn't '\0' `nul`?

zachlefevre:
This will modify what's inserted into truth -- will this also modify what's transmitted to the secondaries?

I assume yes, since you're changing the representation, just want to make sure.

zachlefevre (approved):
Some questions -- but LGTM.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/RepUtils.scala
rjmac: Nope (used to be, but Scala got rid of that, which only really makes sense for C).  Have to use the generalized unicode escape sequence.

rjmac:
> This will modify what's inserted into truth -- will this also modify what's transmitted to the secondaries?

That's actually a really good point.

coordinatorlib/src/main/scala/com/socrata/datacoordinator/common/soql/sqlreps/TextRep.scala
```diff
@@ -52,7 +52,7 @@ class TextRep(val base: String) extends RepUtils with SqlPKableColumnRep[SoQLTyp
 
   def csvifyForInsert(v: SoQLValue) = {
     if(SoQLNull == v) Seq(None)
-    else Seq(Some(v.asInstanceOf[SoQLText].value.replace("\u0000", "")))
+    else Seq(Some(v.asInstanceOf[SoQLText].value))
```
rjmac, line 55: I got rid of this because now it's the Mutator that cleans up input.

zachlefevre (approved):

PR #460, opened 2024-08-26
EN-68632: Try eagerly locking secondary_manifest rows on mutations

rjmac:
In order to hopefully avoid deadlocks based on different operations taking locks in different orders.


saraid (approved):
Worth a try. Too bad we're not likely to be able determine that it worked.

PR #471, opened 2024-11-01
Short circuit collocation explain when all datasets are pre-collocated

rjmac:


housepage (changes requested):
This seems reasonable overall. Can we add a unit tests that tests this scenario is correctly short circuiting?

housepage (approved):
:sheep: :it:

PR #476, opened 2024-12-10
Add a SqlRep for SoQLInterval

rjmac:
This is only a quasi-type; you can't create columns of this type on a dataset (yet) but you _can_ create rollups with columns of this type and therefore it needs a sql rep.  It's not actually used here, only over in soql-postgres-adapter


saraid (approved):
Huh. The translation between three types of interval are kinda weird, but it seems to jive.

PR #478, opened 2025-01-16
removing 'secondary not accepting datasets' check

dmitry-at-tyler:


zachlefevre:
IF you're going to remove the check then you should also remove the config option and the configs in apps-marathon.

dmitry-at-tyler:
> IF you're going to remove the check then you should also remove the config option and the configs in apps-marathon.

Made the tickets, and they will be addressed in a due course at some point in the future.

rjmac (approved):

PR #484, opened 2025-03-11
Ensure the `Memory` implementation actually exists

rjmac:
Not bothering to port over UnsafeMemory; I trust in HotSpot


saraid (approved):
...huh. Okay.

